PR #363, opened 2020-10-07
Add docker and route forwarding for new backend

RomaricMourgues:
The idea is to have everything working in docker transparently.


chamerling (approved):
LGTM

Issue #362, opened 2020-10-07
Add TypeORM in the node backend


chamerling:
1. Create the platform service and integrate https://typeorm.io/#/.
2. Bootstrap some entities in channels service

RomaricMourgues:
Can you also prepare a separated services for the driver ? Do not expose directly the TypeORM ?
I will need:
- a "connect" method that define and start connexion to the database
- a "getDriver" method that returns my modified driver
- a "getRepository" method that returns my modified repository.
Of course you can implement the default TypeORM in this middleware for now.

I also need to change the parameters for the database driver, there is already parameters needed for TypeORM but we'll need more options, so please create a middleware for the driver's option typing.

And remember, we will use MongoDB by default until we have the Cassandra adapter.

chamerling:
Of course, this is what I planned to do.

PR #373, opened 2020-10-09
#362 Add TypeORM as platform service

chamerling:
Note: There is also some work on REST API and best practices with schemas definition and http-errors handling.

### DIFF
--- a/twake/backend/node/README.md
+++ b/twake/backend/node/README.md
@@ -1,1 +1,53 @@
 # Twake backend
+
+## Developer guide
+
+### ORM
+
+We use [TypeORM](https://typeorm.io/) as ORM for all the services of the Twake backend. There are several guidelines to follow in order to manage entities and listeners as described below.
+
+#### Entities
+
+1. Entities have to be defined in folders named `entities`. This comes from the typeorm API which does not allow to register new entities at runtime but allows to define `glob` expression to lookup entities in folders.
+2. Entities must follow the typeorm declaration as defined in the [entity documentation](https://typeorm.io/#/entities).
+
+```js
+import { Entity, Column, ObjectID, ObjectIdColumn } from "typeorm";
+
+@Entity({ name: "channels" })
+export class Channel {
+  @ObjectIdColumn()
+  id: ObjectID;
+
+  @Column()
+  name: string;
+}
+```
+
+#### Subscribers
+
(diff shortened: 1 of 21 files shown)


twake/backend/node/README.md
```diff
@@ -1 +1,53 @@
...
+2. Entities must follow the typeorm declaration as defined in the [entity documentation](https://typeorm.io/#/entities).
+
+```js
+import { Entity, Column, ObjectID, ObjectIdColumn } from "typeorm";
+
+@Entity({ name: "channels" })
+export class Channel {
+  @ObjectIdColumn()
+  id: ObjectID;
+
```
RomaricMourgues, line 21: Can you put an example of usage with something more "Cassandra" ?
Usually we have something like:
"scylladb_keys": {{"workspace_id":"ASC", "parent_id":"ASC", "id":"DESC"}, {"id": "DESC"}} for a primary key composed of ((workspace_id, parent_id), id) and a secondary index with only (id). We will not use secondary indexes anymore, but we will always try to use composed primary keys.
How can we describe this in TypeORM ?

RomaricMourgues: Ok you did it while implementing channels entity, so just update the README.

twake/backend/node/src/services/channels/entities/channel.ts
```diff
@@ -0,0 +1,46 @@
...
+  PUBLIC = "public",
+  DIRECT = "direct"
+}
+
+@Entity({ name: "channels" })
+export class Channel {
+  // uuid-v4
+  @PrimaryColumn()
+  company_id: string;
+
```
RomaricMourgues, line 14: Shouldn't we use a special type for UUID-v4 instead of string ?

chamerling: it depends of the driver used. When really implementing it we will have to take care and it was not the goal of this issue

twake/backend/node/README.md
chamerling: The README is here to show basic samples; we will have to add more when implementation with cassandra is ready

RomaricMourgues:
Ok then tell me when you are ready to merge 👍

chamerling:
@RomaricMourgues this is ready to merge as is. I am looking on the best way to manage entities listeners, it will come in another PR.

PR #378, opened 2020-10-10
🤘🏻 Use RxJS instead of promises to manage service states

chamerling:
This is something I wanted to do when I started to work on the framework. Using RxJS here makes sense to manage component states. This is just the beginning since we want to build something really reactive on the backend.
I will add the unit tests for all of this asap!


RomaricMourgues:
Ok, it could be interesting, let's dont put too many libs right now and see what we need after the first sprint starts. We'll have a better view of what we want to change. But great work anyway!

PR #402, opened 2020-10-13
Feature/2020.q3.dev

RomaricMourgues:


chamerling (commented):
Looks more like some ptrettier/lint fixes than craco integration

RomaricMourgues:
Both, because Craco integrate a linter (Craco is a configuration manager we need for antd).

Issue #425, opened 2020-10-19
Routing implementation


stephanevieira75:
# To do:

## Part 1
- [x] Add react-router-dom package
- [x] Add login, error, setup and client routes
- [x] Update project architecture based on the new routes

## Part 2
- [x] Client router params `web.twake.app/client/workspace-id/c/channel-id`
- [x] Generate shortened id for url
- [ ] Update views while changing routes

PR #443, opened 2020-10-21
 #425 Routing implementation

stephanevieira75:
ref to  #425


twake/frontend/src/app/app.tsx
```diff
@@ -0,0 +1,42 @@
...
+  }, []);
+
+  return (
+    <Integration>
+      <Router history={RouterServices.history}>
+        <Switch>
+          {RouterServices.routes.map((route: any, index: number) => {
+            return (
+              <Route
+                key={route.name + ' ' + index}
```
chamerling, line 22: Avoir string concatenation when possible, use string literals, it makes clean code

twake/frontend/src/app/scenes/Error/Error.tsx
```diff
@@ -0,0 +1,73 @@
...
+    <div className="full_page_error" key="page_error">
+      <div className="error_message skew_in_top_nobounce">
+        <div className="title">
+          <Emojione type=":boom:" /> {Languages.t('scenes.aie', [], 'Aïe !')}
+        </div>
+        <div className="subtitle">
+          {Languages.t(
+            'scenes.error_on_twake',
+            [],
+            'Vous avez trouvé une erreur sur la plateforme Twake !',
```
chamerling, line 28: Why not having them in EN by default?

twake/frontend/src/app/scenes/Error/ErrorBoundary.tsx
```diff
@@ -0,0 +1,40 @@
...
+import RouterServices from '../../services/RouterServices';
+import 'app/ui.scss';
+
+type State = {
+  hasError: boolean;
+  error: any;
+};
+
+export default class ErrorBoundary extends React.Component<{}, State> {
+  constructor(props: any) {
```
chamerling, line 11: I am not a React dev, but can't we have a real type here?

twake/frontend/src/app/scenes/Setup/Setup.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+type ReadyState =
+  | {
+      elasticsearch_connection: boolean;
+      elasticsearch_mapping: boolean;
+      db_connection: boolean;
+      db_mapping: boolean;
+      init: boolean;
+    }
+  | true;
```
chamerling, line 16: Does it means that it is true when every state above is true? You should not define types like this which are changing based on condition. Instead you have to have a `isReady()` function which tells you is it is ready based on the ReadyState.

twake/frontend/src/app/services/Globals.js
```diff
@@ -203,8 +224,8 @@ class Globals {
   }
 
   localStorageGetItem(key, callback) {
-      // eslint-disable-next-line no-undef
-      if (typeof SyncStorage != 'undefined') {
+    // eslint-disable-next-line no-undef
+    if (typeof SyncStorage != 'undefined') {
```
chamerling, line 228: `if (SyncStorage) {` is generally enough

twake/frontend/src/app/services/RouterServices.tsx
```diff
@@ -0,0 +1,66 @@
...
+  routes?: Route[];
+  component?: any;
+  options?: any;
+};
+
+class RouterServices {
+  history = createBrowserHistory();
+
+  // Define your route here
+  pathnames: Readonly<any> = {
```
chamerling, line 23: `Readonly<string>`

twake/frontend/src/app/scenes/Error/Error.tsx
RomaricMourgues: Old component from startup, when all of our clients were french.
You can translate it or remove the fallback Stephane.

twake/frontend/src/app/scenes/Error/ErrorBoundary.tsx
RomaricMourgues: This state can be removed. We don't use this state.

RomaricMourgues: (for the error: any state)
For the (props: any), the type is "{}" like in Component<{}, State> @stephanevieira75 
So (props: {})

twake/frontend/src/app/services/Globals.js
RomaricMourgues: Not in the scope of your issue @stephanevieira75

RomaricMourgues: But thanks for the comment Christophe 👍

Issue #447, opened 2020-10-22
Migration to Ant Design


stephanevieira75:
# To do:

 - [x] Add Ant Design packages
 - [x] Implement Layout antd system for client view 
 - [x] Rename `app` to `Client` 
 - [x] Use typescript for `Client` component

RomaricMourgues:
Antd is already imported (we worked on the style.less I remember no ?)

PR #453, opened 2020-10-22
#447 Add Ant design and implement layout system for client view

stephanevieira75:


twake/frontend/src/app/scenes/Client/app.less
```diff
@@ -1,4 +1,4 @@
-@import '~components/constants.scss';
+@import '~antd/dist/antd.less';
```
RomaricMourgues, line 2: I don't want to use less except for main Antd theme. Can we keep scss file extension ?

twake/frontend/src/app/scenes/Client/WorkspacesBar/WorkspacesBar.js
```diff
@@ -40,7 +41,7 @@ export default class WorkspacesBar extends Component {
     }
 
     return (
-      <div className="workspaces_view fade_in">
+      <Layout.Sider width="60px" className="workspaces_view fade_in">
```
RomaricMourgues, line 44: Can we put the 60px value in the css class workspaces_view ?

RomaricMourgues (changes requested):
Just small changes to do,
very good work !

twake/frontend/src/app/scenes/Client/app.less
stephanevieira75: Done

twake/frontend/src/app/scenes/Client/WorkspacesBar/WorkspacesBar.js
stephanevieira75: We need to keep the width prop or antd design will add a default style :
Capture d’écran de 2020-10-22 17-59-12

stephanevieira75:
> Just small changes to do,
> very good work !

All good for me :smile:

PR #457, opened 2020-10-23
🔌  Collections Transport

RomaricMourgues:
Working on Collections Transport (websockets and http).
- [x] Connecting to websockets and add Collections parameters
- [x] Detect if components are listening to collection or not
- [x] Update docker with socket routes and new backend `npm run dev` command.

Note to myself: there is an issue with WebSQL in multi tab usage (concurrent write on websocket events)


twake/frontend/src/app/services/Observable/Observable.ts
```diff
@@ -49,41 +52,52 @@ export default class Observable {
       const changes = await this.getChanges(watcher);
       if (changes.didChange) {
         //If things changed
-        watcher.callback(changes.changes);
+        watcher.callback(
+          //Could be imporved ?
+          typeof changes.changes === 'object' && changes.changes?.constructor?.name != 'Array'
```
stephanevieira75, line 57: Maybe ? 
`typeof changes.changes === 'object' && !changes.changes instanceof Array`

(I'm not sure about this)

PR #461, opened 2020-10-23
#425 Update client routes and allow url parameters

stephanevieira75:
## Default client route for channel, applications and private message

Capture d’écran de 2020-10-23 17-31-57


twake/frontend/src/app/services/RouterServices.tsx
```diff
@@ -23,19 +23,20 @@ class RouterServices {
   history = createBrowserHistory();
 
   // Define your route here
-  pathnames: Readonly<Pathnames> = {
+  public pathnames: Readonly<Pathnames> = {
     CLIENT: '/client',
+    CLIENT_APP: '/client/w/:workspace/c/:channel',
```
chamerling, line 28: do we really need `/client` prefix?

RomaricMourgues: "client" yes but "w" not @stephanevieira75 :) 
We stole it from Slack, it'll be needed in the future to have a constant prefix (we wanted a smaller prefix, but web.twake.app/app or web.twake.app/web seems a bit redundant !)

stephanevieira75: Removed :+1:

PR #472, opened 2020-10-27
WIP Work on cassandra + node

tissarni:


twake/backend/node/src/core/platform/services/database/services/connectors/cassandra.ts
```diff
@@ -22,7 +24,10 @@ export class CassandraConnector extends AbstractConnector<CassandraConnectionOpt
 
     const cassandraOptions: cassandra.DseClientOptions = {
       contactPoints: this.options.contactPoints,
+      localDataCenter: 'datacenter1',
```
chamerling, line 27: Should come from config file

twake/backend/node/src/services/channels/services/index.ts
```diff
@@ -10,11 +10,11 @@ import { CassandraChannelService } from "./cassandra";
 export function getService(databaseService: DatabaseServiceAPI): ChannelServiceAPI<Channel> {
   const type = databaseService.getConnector().getType();
 
-  switch(type) {
+  switch (type) {
     case "mongodb":
       return new MongoChannelService((databaseService.getConnector() as MongoConnector).getDatabase());
     case "cassandra":
-      throw new CassandraChannelService((databaseService.getConnector() as CassandraConnector).getClient());
+      return new CassandraChannelService((databaseService.getConnector() as CassandraConnector).getClient());
```
chamerling, line 17: Sorry for this, hope you did not loose time on it

twake/backend/node/src/services/channels/web/controller.ts
```diff
@@ -2,14 +2,28 @@ import { Channel } from "../entities";
...
+
 export default class ChannelController {
-  constructor(private service: ChannelServiceAPI<Channel>) {}
+  constructor(private service: ChannelServiceAPI<Channel>) { }
 
   async create(channel: CreateChannelBody): Promise<Channel> {
     const entity = new Channel();
     // TODO: Create the Channel entity from the CreateChannelBody
     // The CreateChannelBody is already validated by the web framework
+
```
chamerling, line 13: I introduced class-transformer in some part of my code, going to push a PR so that we can reuse it and avoid this

twake/backend/node/src/services/channels/web/schemas.ts
```diff
@@ -2,34 +2,105 @@ export const createChannelSchema = {
   body: {
     type: "object",
     properties: {
-      creator: {
-        type: "string",
+      resource: {
+        type: "object",
```
chamerling, line 6: OK, going to push something to be able to share schemas so that we do not dup them

tissarni:
My work on Cassandra CRUD with one SAVE route. The save method in case of creation is not finished but can work (I have to define in the url the id of the channel instead of automatically generating an id) and the getAll method does not work because I created a wrong table (bad partition key / clustering key / primary key)

twake/backend/node/src/core/platform/framework/api/crud-service.ts
```diff
@@ -43,12 +30,11 @@ export class DeleteResult<Entity> extends EntityTarget<Entity>  {
...
-export declare type EntityOperationResult<Entity> = CreateResult<Entity> | UpdateResult<Entity> | DeleteResult<Entity>;
+export declare type EntityOperationResult<Entity> = SaveResult<Entity> | DeleteResult<Entity>;
 
 export interface CRUDService<Entity> {
-  create(item: Entity): Promise<CreateResult<Entity>>;
-  get(id: EntityId): Promise<Entity>;
-  update(id: EntityId, item: Entity, /* TODO: Options */): Promise<UpdateResult<Entity>>;
-  delete(id: EntityId): Promise<DeleteResult<Entity>>;
-  list(/* TODO: Options */): Promise<Entity[]>;
+  save(item: Entity): Promise<SaveResult<Entity>>;
```
chamerling, line 36: OK for the primary key, but it must be a generic so that we can choose it based on the type of database:

```js
export interface CRUDService<Entity, PrimaryKey> {
   get(primaryKey: PrimaryKey): Promise<Entity>;
}
```

twake/backend/node/src/core/platform/framework/api/crud-service.ts
```diff
@@ -43,12 +30,11 @@ export class DeleteResult<Entity> extends EntityTarget<Entity>  {
...
+  get(pk: { [column: string]: EntityId }): Promise<Entity>;
+  delete(pk: { [column: string]: EntityId }): Promise<DeleteResult<Entity>>;
+  list(pk: { [column: string]: EntityId }/* TODO: Options */): Promise<Entity[]>;
```
chamerling, line 39: Not sure we have pk in list. We will have options soon.

twake/backend/node/src/core/platform/framework/decorators/realtime/index.ts
```diff
@@ -1,14 +1,13 @@
...
+export * from "./saved";
 export * from "./deleted";
-export * from "./updated";
 
 export interface PathResolver<T> {
   (type: T): string;
 }
 
-export function getPath<T>(path : string | PathResolver<T> = "/", entity: EntityTarget<T>): string {
+export function getPath<T>(path: string | PathResolver<T> = "/", entity: EntityTarget<T>): string {
```
chamerling, line 10: Avoid such changes which are useless in this scope. We will add prettier stuff and make a single commit for all the fixes

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -2,34 +2,97 @@
 import cassandra from "cassandra-driver";
 import { Channel } from "../entities";
 import ChannelServiceAPI from "../provider";
-import { RealtimeDeleted, RealtimeCreated, RealtimeUpdated } from "../../../core/platform/framework/decorators";
-import { UpdateResult, CreateResult, DeleteResult } from "../../../core/platform/framework/api/crud-service";
+import { RealtimeDeleted, RealtimeSaved } from "../../../core/platform/framework/decorators";
+import { SaveResult, DeleteResult } from "../../../core/platform/framework/api/crud-service";
+import { resolve } from "path";
```
chamerling, line 7: useless import

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -2,34 +2,97 @@
...
+
   }
 
   @RealtimeDeleted<Channel>("/channels", channel => `/channels/${channel.id}`)
-  async delete(id: string): Promise<DeleteResult<Channel>> {
-    throw new Error("Not implemented");
+  async delete(pk: { [column: string]: string | number }): Promise<DeleteResult<Channel>> {
+    const channelToDelete = await this.get(pk);
+    const query = 'DELETE FROM twakechannel.channel_data WHERE id = ? AND company_id = ? AND workspace_id = ? ';
+    const params = [pk.id, pk.company_id, pk.workspace_id];
```
chamerling, line 61: avoid creating variables for nothing, in this case, you can simply do 

```js
    await this.client.execute(query, [pk.id, pk.company_id, pk.workspace_id]);
```

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -2,34 +2,97 @@
...
+    await this.client.execute(query, params);
+
+    const result = new DeleteResult<Channel>('channel', channelToDelete, true);
```
chamerling, line 64: Return result directly

chamerling: Are we sure that the channel has been deleted? If the client result can provide the information please use it

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -2,34 +2,97 @@
...
   }
 
-  async list(/* TODO: Options */): Promise<Channel[]> {
-    throw new Error("Not implemented");
+  async list(pk: { [column: string]: string | number }/*TO DO*/): Promise<Channel[]> {
+    const query = 'SELECT * FROM twakechannel.channel_data WHERE company_id = ? AND workspace_id = ?';
+    const params = [pk.company_id, pk.workspace_id];
+    let resultArray: Channel[] = [];
+
+    await this.client.execute(query, params)
```
chamerling, line 73: If you await, you do not have to do `then()`:

```js
const results = await this.client.execute(query, params);
```

and then 

```js
return results.rows.map(row => ({
  company_id: row.values()[0],
  workspace_id: row.values()[1],
  // etc
});

File: twake/backend/node/src/services/channels/web/routes.ts [hunk 12]
```diff
@@ -59,18 +59,18 @@ const routes: FastifyPluginCallback<{ service: ChannelServiceAPI<Channel> }> = (
     }
   });
 
-  fastify.route<{ Body: CreateChannelBody, Params: ChannelParameters }>({
+  fastify.route<{ Body: { resource: CreateChannelBody }, Params: ChannelParameters }>({
     method: "POST",
-    url,
+    url: `${url}/:id`,
```
chamerling, line 64: This breaks REST. If you need to define an id, do it in the HTTP body

twake/backend/node/src/services/channels/web/routes.ts
```diff
@@ -59,18 +59,18 @@ const routes: FastifyPluginCallback<{ service: ChannelServiceAPI<Channel> }> = (
...
-      request.log.debug(`Creating Channel ${JSON.stringify(request.body)}`);
-
-      const resource = await controller.create(request.params, request.body);
+      request.log.debug(`Save Channel ${JSON.stringify(request.body)}`);
+      const resource = await controller.save(request.params, request.body.resource);
 
       if (resource) {
-        reply.code(201);
+        console.log(request.params, request.body);
+        reply.code(200);
```
chamerling, line 73: When a resource is created, we return 201.

twake/backend/node/src/core/platform/framework/decorators/realtime/index.ts
RomaricMourgues: We already all use prettier, I think you are the only one not using it @chamerling ;)

twake/backend/node/src/services/channels/web/routes.ts
RomaricMourgues: This is because there was only one route for POST, and I was thinking you implemented a single "save" route so I told Titouan to fix it like this. We need to discuss tomorrow about this route thing

twake/backend/node/src/services/channels/web/routes.ts
RomaricMourgues: And when updated ?

twake/backend/node/src/services/channels/web/routes.ts
RomaricMourgues: And by the way the more we do REST the more I find it old school. We'll continue using it of course, too late to change,  but I think POST everywhere + only one endpoint is better.
Why having data in headers, url, query string and body? Isn't it crasy in the end?
I like the http code response, but why in header?
Notion is working this way for instance.

Anyway this is just my through of this evening :)

twake/backend/node/src/core/platform/framework/decorators/realtime/saved.ts
```diff
@@ -8,20 +8,20 @@ import { eventBus } from "../../../services/realtime/bus";
  * @param path the path to push the notification to
  * @param resourcePath the path of the resource itself
  */
-export function RealtimeUpdated<T>(path: string | PathResolver<T>, resourcePath?: string | PathResolver<T>): MethodDecorator {
+export function RealtimeSaved<T>(path: string | PathResolver<T>, resourcePath?: string | PathResolver<T>): MethodDecorator {
   // eslint-disable-next-line @typescript-eslint/ban-types
-  return function(target: Object, propertyKey: string, descriptor: PropertyDescriptor ): void {
+  return function (target: Object, propertyKey: string, descriptor: PropertyDescriptor): void {
```
stephanevieira75, line 13: `function (target: Object,...)` or ` function (target: object,...)` ?
I think you should use the second one in typescript.

twake/backend/node/src/services/channels/web/types.ts
```diff
@@ -51,4 +51,10 @@ export interface ChannelWebsocket {
 export class CreateChannelBody {
```
stephanevieira75, line 51: All string types ?
Maybe you could something like ` [key: string]: string;`.

WDYT @chamerling  ?

twake/backend/node/test/http/channels.http
```diff
@@ -1,7 +1,7 @@
 @company_id = 0
 @workspace_id = 0
-@baseURL = http://localhost:3000/api
-@channelsURL = {{baseURL}}/channels/companies/{{company_id}}/workspaces/{{workspace_id}}
+@baseURL = http://localhost:8000/internal/services/channels/v1
+@channelsURL = {{baseURL}}/companies/bcfe2f79-8e81-42a3-b551-3a32d49b2b4c/workspaces/3328552c-5ccd-4172-b84a-d876d56aa71b
```
stephanevieira75, line 4: Hard coded ids, this is intended ?

twake/backend/node/package.json
```diff
@@ -70,6 +71,7 @@
     "cassandra-driver": "^4.6.0",
     "class-transformer": "^0.3.1",
     "config": "^3.3.2",
+    "express-cassandra": "^2.5.0",
```
chamerling, line 74: useless

twake/backend/node/src/core/platform/framework/api/crud-service.ts
```diff
@@ -4,23 +4,10 @@ export class EntityTarget<Entity> {
...
    */
-  constructor(readonly type: string, readonly entity: Entity) {}
+  constructor(readonly type: string, readonly entity: Entity) { }
 }
 
-export class UpdateResult<Entity> extends EntityTarget<Entity> {
-  /**
-   * Result sent back by the underlying database
-   */
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
chamerling, line 14: I will be happy to keep this in the SaveResult

twake/backend/node/src/core/platform/framework/api/crud-service.ts
chamerling: Did you have time to work on this? If not I will do it next week, focus on other parts

twake/backend/node/src/core/platform/framework/decorators/realtime/index.ts
chamerling: VSCode is not happy with so much projects and config for now, will have to find the way to configure it correctly

twake/backend/node/src/core/platform/framework/decorators/realtime/saved.ts
chamerling: We need to define the rules for this

twake/backend/node/src/core/platform/services/database/services/connectors/cassandra.ts
```diff
@@ -2,6 +2,7 @@
 import cassandra from "cassandra-driver";
 import { AbstractConnector } from "./abstract-connector";
 
+
```
chamerling, line 5: Please avoid to add blank lines for nothing. A common rule is to have 1 blank line

twake/backend/node/src/services/channels/entities/channel.ts
```diff
@@ -1,9 +1,10 @@
 import { ObjectID } from "mongodb";
+import Uuid from "cassandra-driver"
```
chamerling, line 2: No, we must not rely on cassandra-driver in this place. If we have to, this is because the node-uuid module does not provide what we need

twake/backend/node/src/services/channels/index.ts
```diff
@@ -6,7 +6,7 @@ import web from "./web/index";
 import { Channel } from "./entities";
 import { DatabaseServiceAPI } from "../../core/platform/services/database/api";
 
-@Prefix("/api/channels")
+@Prefix("/internal/services/channels/v1")
```
chamerling, line 9: I still do not understand why we need to change this. We can have this API at http://localhost:3000/api/channels. This is up to the reverse proxy to redirect http://twake:port/internal/services/channels/v1 to http://localhost:3000/api/channels

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -1,35 +1,103 @@
...
+        result.icon = row.values()[8];
+        result.name = row.values()[9];
+        result.description = row.values()[7];
+        result.channel_group = row.values()[5];
+        result.visibility = row.values()[11];
+        result.is_default = row.values()[6];
+        result.archived = row.values()[4];
+        result.archivation_date = row.values()[3];
+      });
+    return result;
```
chamerling, line 61:
```js
const row = await this.client.execute(/*your params*/).first();
```

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -1,35 +1,103 @@
...
   }
 
   @RealtimeDeleted<Channel>("/channels", channel => `/channels/${channel.id}`)
-  async delete(id: string): Promise<DeleteResult<Channel>> {
-    throw new Error("Not implemented");
+  async delete(pk: { [column: string]: string | number }): Promise<DeleteResult<Channel>> {
+    const channelToDelete = await this.get(pk);
```
chamerling, line 66: Why do you need to do this before deleting?

twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -1,35 +1,103 @@
...
+    const query = 'DELETE FROM twakechannel.channel_test WHERE id = ? AND company_id = ? AND workspace_id = ?';
+    //const params = [pk.id, pk.company_id, pk.workspace_id];
+    await this.client.execute(query, [pk.id, pk.company_id, pk.workspace_id]);
+
+    return (new DeleteResult<Channel>('channel', channelToDelete, true))
```
chamerling, line 71: Can't we have the delete result from the driver to know if we really deleted the channel?

twake/backend/node/src/services/channels/web/routes.ts
chamerling: 200 with the updated object

twake/backend/node/src/core/platform/framework/decorators/realtime/saved.ts
RomaricMourgues: Simply install prettier on VSCode and copy the .prettierrc from twake/frontend/.prettierrc
Then enable prettify on save.

RomaricMourgues:
Retrieved by Christophe on its side

PR #490, opened 2020-10-29
Run Node backend tests in docker

chamerling:
I know that this is not the current way to do it, I did not wanted to change it, just to show how it should work in the best of the worlds:

1. Each runtime should have its `Dockerfile`
2. A docker-compose file is used for tests to launch test platform and all tests by default
3. The config/test.json file override the mongo configuration so that the node container can connect to the mongo one using `mongo` as host name
4. User can launch by hand, cf README
5. The `Dockerfile` has multiple stages, for test, development and production. It uses a light node container

If we agree on this, I will update the github workflow. Anyway, the current step `- run: cd twake && docker-compose exec -T node npm run test` fails just because the node container is not started: We should not start it in a compose and try to use it in another one like it is currently done.


RomaricMourgues:
Your docker-compose is clean but it will quickly contain cassandra, elasticsearch, rabbitmq and php, why not keeping our current dockerfile here? https://github.com/TwakeApp/Twake/blob/feature/2020.Q4.dev/twake/docker-compose.yml.dist.tests
Currently testing Twake is done like this:
https://github.com/TwakeApp/Twake/blob/feature/2020.Q4.dev/.github/workflows/backend.yml

Why not testing Twake like in production?
Cons: slow to test as a developer
Pros: we are sure that it will work when installing full Twake

The issue you point (the node container is not started) is an issue with our docker-compose so it must be fixed no?
I would like to have a test that simulate what a community member would do to run Twake (I mean the docker-compose up)

In the future you'll manage this @aduprat so I' like to ear your point of view :)

aduprat (approved):
I agree with this first step.

I think that middleware lifecycle should be handled by tests themselves.
I'm pretty sure that this library is a good candidate for that:
https://github.com/testcontainers/testcontainers-node

RomaricMourgues:
If you both agree then Christophe you can update the workflow.

In this case we need to have the same environment in production than in tests also, so please remove this folder and do the needed changes to each docker-compose.yml :
https://github.com/TwakeApp/Twake/tree/feature/2020.Q4.dev/twake/docker/twake-node

chamerling:
Allright, will update things asap

RomaricMourgues:
Is it ready for merge or do you want to fix Cassandra testing first?

RomaricMourgues (approved):

PR #505, opened 2020-10-30
↔️ Use websocket details for transport

RomaricMourgues:
- Use the websocket details from GET calls to initialise websockets
- Auto-rejoin websocket rooms on disconnection
- Offline mode with IndexedDb (fallback to MemoryDB) 
- If no configuration, no http request are done
- Apply linter to Collections and Observables

### DIFF
--- a/twake/frontend/src/app/services/Collections/Transport/TransportSocket.ts
+++ b/twake/frontend/src/app/services/Collections/Transport/TransportSocket.ts
@@ -1,14 +1,30 @@
-import Collections, { Collection, Resource } from '../Collections';
+import Collections from '../Collections';
 import Transport from './Transport';
 import io from 'socket.io-client';
 
+export enum WebsocketActions {
+  Join = 'realtime:join',
+  Leave = 'realtime:leave',
+}
+
+export enum WebsocketEvents {
+  Connected = 'connected',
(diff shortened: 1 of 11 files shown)


twake/frontend/src/app/services/Collections/Transport/TransportSocket.ts
```diff
@@ -21,13 +32,28 @@ export default class TransportSocket {
         .on('unauthorized', (err: any) => {
           console.log('Unauthorize', err);
         });
+
+      Object.values(this.listeners).forEach(element => {
+        element(WebsocketEvents.Connected, {});
+      });
+
+      socket.on('realtime:join:success', (event: any) => {
```
chamerling, line 40: Can use `WebsocketEvents.JoinSuccess`, same comment for the ones below

twake/frontend/src/app/services/Collections/Transport/TransportSocket.ts
```diff
@@ -21,13 +32,28 @@ export default class TransportSocket {
...
+        this.notify(event?.name, WebsocketEvents.JoinSuccess, event);
```
chamerling, line 41: Don't you have error since `notify` path is not optional?

PR #510, opened 2020-10-31
Update docker-compose.yml.dist

objeni:
can we use other port number of twake


RomaricMourgues:
Hi, yes you can for yourself, but why do you want to change the port for everyone?
Change it on your side and then do `docker-compose stop; docker-compose up -d;` no need to change it on the official repository :)

objeni:
but it not working after changing the port I just testing that it works or not because if some other applications uses the same port then we change that port or not.

PR #514, opened 2020-11-02
Add the channel ressource

RkAiman:


RomaricMourgues:
Note: currently all fields are optional, we'll get back to this later to improve the Collections framework.

PR #515, opened 2020-11-02
#513 cassandra channels

chamerling:


twake/backend/node/src/services/channels/services/cassandra.ts
```diff
@@ -1,35 +1,82 @@
...
+      "workspace_id",
+      "id",
+      "icon",
+      "name",
+      "description",
+      "channel_group",
+      "visibility",
+      "is_default",
+      "archived"
+      )
```
RomaricMourgues, line 33: Can we map over parameters of the entity in the future here?

chamerling:
Yes this is the idea. With some decorators it will be automatic and so we
will avoid writing queries.

Le lun. 2 nov. 2020 à 18:31, Romaric Mourgues <notifications@github.com> a
écrit :

twake/backend/node/src/services/channels/services/cassandra.ts
chamerling: Yes this is the idea. With some decorators it will be automatic and so we
will avoid writing queries.

twake/backend/node/src/services/channels/entities/channel.ts
```diff
@@ -8,13 +9,18 @@ export enum VisibilityEnum {
...
   // uuid-v4
+  @Type(() => String)
   company_id: string;
 
   // "uuid-v4" | "direct"
+  @Type(() => String)
   workspace_id: string;
 
+  @Type(() => String)
+  @Exclude({ toPlainOnly: true })
```
RomaricMourgues, line 20: Do you need to declare _id in the entities as this is something
- that must not be sent to the frontend
- that does not exists in cassandra
?

twake/backend/node/src/services/channels/provider.ts
```diff
@@ -2,4 +2,12 @@ import { CRUDService } from "../../core/platform/framework/api/crud-service";
 import { TwakeServiceProvider } from "../../core/platform/framework/api";
 import { Channel } from "./entities";
 
-export default interface ChannelServiceAPI extends TwakeServiceProvider, CRUDService<Channel> {}
+export type ChannelPrimaryKey = {
+  id?: string;
+  company_id?: string;
+  workspace_id?: string;
+};
```
RomaricMourgues, line 9: How do you define the primary key order ?
In Cassandra the primary key contains a partition key and clustering keys. For instance in this case we have:
Partition key = [company_id]
Clustering key = [workspace_id, id]
In cassandra this is defined like this: ((company_id), workspace_id, id).

>It means that we cannot access the resources using:
company_id and id,
workspace_id and id,
id alone
workspace_id alone
>We only can access resources using:
company_id, company_id and workspace_id, company_id and workspace_id and id

The clustering keys also contain sorting information.

So we need to be able to define all this information :)

RomaricMourgues (approved):
I approve the merge request but I added some comments.

twake/backend/node/src/services/channels/entities/channel.ts
chamerling: I need to check how to handle it in the schema, in mongo and in cassandra to only use `id`.
This will be part of some other PR.

twake/backend/node/src/services/channels/provider.ts
chamerling: Just tell me how to do this. I guess this is defined when we create the table, and then when we build the queries.

twake/backend/node/src/services/channels/entities/channel.ts
chamerling: OK got it, pushing something within the next minutes 🍻

chamerling: Here it is 🍺

twake/backend/node/src/services/channels/provider.ts
RomaricMourgues: Yes, I'm used to have the table generation done by the ORM.
Could we have this information here anyway even if node will not generate the tables ?
Something like:
```
@Entity(() => {
    table: "channel",
    primaryKey: [["company_id"], "workspace_id", "id"]
})
export class Channel {
```

chamerling: Yes, we will but not in this PR, it starts to be huge. I will check how we can automate this, if we can. If not we will provide scripts.

PR #576, opened 2020-11-17
WIP 🔪 Kill SocketCluster

RomaricMourgues:
Removing socket cluster replacing it by socket.io from node backend.


RomaricMourgues:
Merging this, we'll se in an other PR how to improve communication using maybe rabbit or something like this.

Issue #555, opened 2020-11-12
Channel edition / creation popup


RomaricMourgues:
We need to change the channel editor popup to make it working for channel edition following the US:

- [ ] Display visibility edition only if I am the owner or a workspace administrator (need a clean service to get user right)

- [ ] Do not open channel member edition popup in edition

PR #593, opened 2020-11-26
#555 Add channel edition for the owner

stephanevieira75:


twake/frontend/src/app/scenes/Client/ChannelsBar/Parts/Channel/ChannelMenu.tsx
```diff
@@ -185,6 +185,11 @@ export default (props: Props): JSX.Element => {
         },
       },
     );
+
+    // This remove the edit channel menu item when user isn't channel owner
+    if (props.channel.owner !== currentUser.id) {
+      menu.splice(4, 1);
```
chamerling, line 191: Is it the way to do in react, seriously? Can't we apply rules on each menu item based on the user roles?

stephanevieira75: Something like this ?

```js
{
      type: 'menu',
      rules: props.channel.owner === currentUser.id,
      text: Languages.t('scenes.app.channelsbar.modify_channel_menu'),
      onClick: () => {
        editChannel();
      },
}
...

const filteredMenu = menu.filter((item: any) => item.rules !== false);
```

RomaricMourgues: We could add if / else while pushing menu elements, but your proposition @stephanevieira75 is good, I really like it.
Can you:
- rename "rules" to "disabled" (and so inverse the condition)
- put this line 175 of MenusBodyLayer:
```
          {this.state.menus_manager.menus.filter(item=>!item.disabled).map((item, i) => {
```

chamerling: Well, item can be enabled/disabled/hidden, so is `disabled` the right name?

RomaricMourgues: We already had a "hide" option in the menu builder, so we used it 👍

PR #607, opened 2020-11-30
Direct channel creation

chamerling:
TODO:

- [x] Block update
- [x] Always add current user in members list
- [x] Fix visibility in channel when workspace is direct: Assign visibility from workspace
- [x] Publish in websocket

### DIFF
--- a/twake/backend/node/src/services/channels/services/channel/cassandra.ts
+++ b/twake/backend/node/src/services/channels/services/channel/cassandra.ts
@@ -175,9 +175,6 @@ export class CassandraChannelService implements ChannelService {
     const query = `INSERT INTO ${this.options.keyspace}.${this.table} (${columnList}) VALUES (${columnValues})`;
 
     logger.info("service.channel.create - %s - %o", query, saveChannel);
-    console.log(typeof channel.company_id);
-    console.log(typeof channel.workspace_id);
-    console.log(typeof channel.id);
 
     await this.client.execute(query, saveChannel, { prepare: true });
 
@@ -257,7 +254,7 @@ export class CassandraChannelService implements ChannelService {
     const query = `SELECT * FROM ${this.options.keyspace}.${this.directChannelsTableName} WHERE company_id = ? AND channel_id = ? AND users = ?;`;
     const result = await this.client.execute(query, directChannel);
 
-    if (!result.rows) {
+    if (!result.rowLength) {
       return;
     }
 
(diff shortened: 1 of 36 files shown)


twake/backend/node/src/services/channels/services/channel/cassandra.ts
```diff
@@ -55,27 +57,76 @@ export class CassandraChannelService implements ChannelService {
...
+  }
+
+  protected async createDirectChannelTable(): Promise<boolean> {
+    const query = `
+      CREATE TABLE IF NOT EXISTS ${this.options.keyspace}.${this.directChannelsTableName}
+        (
+          company_id uuid,
+          channel_id uuid,
+          users text,
+          PRIMARY KEY ((company_id, channel_id, users))
```
RomaricMourgues, line 101: If this table is the "DIRECT_CHANNEL_IDENTIFIERS" from documentation (used to find the corresponding channel for a list of users) then you must have ((company_id), users) as primary key.

Primary key = ((partition key), clustering keys)

Some examples:
PK: ((a), b, c)
=> ✅ WHERE a
=> ✅ WHERE a and b
=> ✅ WHERE a and b and c
=> ⛔️ WHERE a and c
=> ⛔️ WHERE c
=> ⛔️ WHERE b and c

PK: ((a, b), c) (the Partition key is the two first one grouped)
=> ⛔️ WHERE a
=> ✅ WHERE a and b
=> ✅ WHERE a and b and c
=> ✅ WHERE a and b ORDER BY c ASC
=> ✅ WHERE a = ... and b = ... and c > ...
=> ✅ WHERE a = ... and b = ... and c < ...
=> ⛔️ WHERE a = ... and b < ... (cannot order on a partition key)

PK: ((a), b ASC, c ASC)
=> ✅  WHERE a = ... and b = ... and c < ...
=> ✅  WHERE a = ... and b < ... and c < ...
=> ✅  WHERE a = ... and b > ... and c > ...
=> ⛔️  WHERE a = ... and b > ... and c < ... (cannot mix orders, reverse all or use as defined but no other options)

Note: the partition key is to define where is the data in the datacenter, so the best is to have one node for one partition key related data. But too much partition key is not good eithier. Thinking that "one company = one node" can be a solution, and then always use company_id as partition key and other keys as clustering keys.
Of course if we think it could be too many items in a single partition key, then we can add more (like workpsace id)

In conclusion, use ((company_id, workspace_id), other keys) or ((company_id), other keys) if workspace_id does't exists.

RomaricMourgues: So you must change also the PK of the channelMembersTable and userChannelsTable to reduce size of Partition Key.

RomaricMourgues (approved):

Issue #606, opened 2020-11-30
Add options to tabs (rename, reset configuration, delete)


RomaricMourgues:
Add options to tabs, rename is available to everyone, but remove and configurate is only available to:
- the tab owner (to add in the tab model)
- the channel owner
- administrators

PR #612, opened 2020-12-02
#606 Improve tabs design & allow renaming or deleting a tab based on user roles

stephanevieira75:
# Preview:
Twake (1)

last commit fix this:

Capture d’écran de 2020-12-02 17-51-46

### DIFF
--- a/twake/frontend/src/app/scenes/Client/MainView/Tabs/Tab.tsx
+++ b/twake/frontend/src/app/scenes/Client/MainView/Tabs/Tab.tsx
@@ -13,6 +13,7 @@ import { capitalize } from 'lodash';
 import AccessRightsService from 'app/services/AccessRightsService';
 import Collections from 'app/services/Collections/Collections';
 import { ChannelResource } from 'app/models/Channel';
+import { Typography } from 'antd';
 
 type PropsType = {
   tabResource: TabResource;
@@ -59,7 +60,16 @@ export default ({ tabResource, upsertTab, deleteTab, currentUserId }: PropsType)
       }}
     >
       {getAppIcon(tabResource)}
-      <span className="small-right-margin">{capitalize(tabResource.data.name)}</span>
+      <Typography.Paragraph
+        ellipsis={{
+          rows: 1,
+          expandable: false,
+        }}
+        style={{ maxWidth: '108px', marginBottom: 0 }}
+        className="tab-name small-right-margin"
+      >
+        {capitalize(tabResource.data.name)}
+      </Typography.Paragraph>
       {tabResource.data.id === tabId && (
         <Menu
           style={{ lineHeight: 0 }}
@@ -84,7 +94,7 @@ export default ({ tabResource, upsertTab, deleteTab, currentUserId }: PropsType)
               type: 'menu',
               hide:
                 currentUserId !== tabResource.data.owner &&
-                isCurrentUserAdmin === false /*&&
+                !isCurrentUserAdmin /*&&
                 isChannelOwner()*/,
               text: <div style={{ color: 'var(--red)' }}>{Languages.t('general.delete')}</div>,
               onClick: () => deleteTab(tabResource),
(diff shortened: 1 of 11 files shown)


twake/frontend/src/app/scenes/Client/MainView/Tabs/TabsTemplateEditor.tsx
```diff
@@ -32,21 +33,43 @@ export default (props: PropsType): JSX.Element => {
...
+                configuration: {},
+              });
+            }
+
+            return props.onChangeTabs(editedTab);
           }}
         >
-          Create
+          {Languages.t(props.tab?.data.id ? 'general.edit' : 'general.create')}
         </Button>
```
RomaricMourgues, line 73: We must not be able to change the app in edition (so hide the select in edition)

RomaricMourgues: If you could also set the name input and select input sizes to large:  <Select size={'large'}  <Input size={'large'}

RomaricMourgues (changes requested):

twake/frontend/src/app/services/workspaces/workspaces_apps.js
```diff
@@ -292,16 +292,19 @@ class WorkspacesApps extends Observable {
...
-    if (app.simple_name.toLocaleLowerCase() == 'twake_calendar') {
-      return feather ? Calendar : 'calendar-alt';
-    }
-    if (app.simple_name.toLocaleLowerCase() == 'twake_drive') {
-      return feather ? Folder : 'folder';
-    }
-    if (app.simple_name.toLocaleLowerCase() == 'twake_tasks') {
-      return feather ? CheckSquare : 'check-square';
+    if (app) {
+      if (app.simple_name.toLocaleLowerCase() == 'twake_calendar') {
```
chamerling, line 296: Even if it can be done with a switch, if you want to keep so many if, simplify it like:

```js
const appName = app.simple_name. toLocaleLowerCase();

if (appName === 'twake_calendar') {
  return return feather ? Calendar : 'calendar-alt';
}
// etc
```

twake/frontend/src/app/scenes/Client/MainView/Tabs/Tab.tsx
```diff
@@ -0,0 +1,99 @@
...
+                    position: 'center',
+                    size: { width: '500px', minHeight: '329px' },
+                  },
+                ),
+            },
+            {
+              type: 'menu',
+              hide:
+                currentUserId !== tabResource.data.owner &&
+                isCurrentUserAdmin === false /*&&
```
chamerling, line 87: `&& !isCurrentUserAdmin`

chamerling (commented):
1. What happens if the tab name is lonnnnnnnnngggggggg. Do we have a limit and an ellipsis?
2. Can we drag and drop the tabs to change order? (I guess no, but will be nice to, maybe later)

twake/frontend/src/app/services/workspaces/workspaces_apps.js
RomaricMourgues: > 1. What happens if the tab name is lonnnnnnnnngggggggg. Do we have a limit and an ellipsis?
> 2. Can we drag and drop the tabs to change order? (I guess no, but will be nice to, maybe later)

1. I am for limiting size on tab names @stephanevieira75 
2. We put an "order" field in the tabs resource model, so we will be able to do it in the future! But not now, we could have a "move left" "move right" button but I don't want to think about drag-drop so soon, may be next year :)

twake/frontend/src/app/scenes/Client/MainView/Tabs/Tab.tsx
stephanevieira75: Done

twake/frontend/src/app/services/workspaces/workspaces_apps.js
stephanevieira75: Updated the getAppIcon function, added an ellipsis for the tab name and  max length for the input. :smiley:

twake/frontend/src/app/scenes/Client/MainView/Tabs/TabsTemplateEditor.tsx
stephanevieira75: Done :+1:

PR #637, opened 2020-12-07
Cassandra/Mongo ORM

RomaricMourgues:


twake/backend/node/src/services/channels/services/tab/service.ts
```diff
@@ -9,78 +9,104 @@ import {
...
   async init(): Promise<this> {
-    try {
-      this.service.init && (await this.service.init());
-    } catch (err) {
-      console.error("Can not initialize channel tab service");
-    }
+    await this.database.getRepository("channel_tab").init(ChannelTab);
+    const manager = this.database.newManager();
+
+    const tab = new ChannelTab();
```
chamerling: Great, but you should remove this which seems to be tests :)

chamerling (approved):
I ❤️  it! Thanks for this

Issue #658, opened 2020-12-09
Create the pubsub link between channel service preferences and notification service one


chamerling:
When a member join a channel, updates his preferences, leave a channel, the channel service will push notification into the pubsub service. On the other side, the notification service subscribes to these notifications and creates/updates/deletes notification service level entities

PR #661, opened 2020-12-09
#658 Bootstrap pubsub between channels and notifications

chamerling:


chamerling:
@RomaricMourgues can you check the commit d553126, I changed some little stuff on the ORM side

chamerling:
> Your proposal is very tempting, and no far to reach, we just need to use the entity decorator to make the class extends some basic Entity functions like save and remove. But I don't see how we can do it without being too much of a black box (if we want to keep a buffer).
> Anyway, let continue with this for now, feel free to improve it as soon as we don't take too much time in it and we don't do too many breaking changes ;)

I did some changes tonight, not changed lot of things but trying to clean a little bit. You can still get a manager reference from the API if you need to do bulk. If not, you can directly work on the repository without having to flush things yourself, etc... I will split files and move things around tomorrow

RomaricMourgues (approved):

Issue #227, opened 2020-09-04
Indicator when there is a notification from a hidden channel / contact  


BenoitTallandier:
**Is your feature request related to a problem? Please describe.**
When a notification arrives on a hidden channel, we don't know about it if we don't scroll.

**Describe the solution you'd like**
Indicator at the bottom (or top) of the interface when there is a notification of a hidden channel / contact

RomaricMourgues:
Duplicate of #178 (I close #178 because this one is better defined)

BenoitTallandier:
Related to #307

PR #680, opened 2020-12-14
 #227 Add indicator when there is a notification from a hidden channel / contact 

stephanevieira75:
# Preview

Twake (2)
Twake (3)

### DIFF
--- a/twake/frontend/src/app/scenes/Client/ChannelsBar/ChannelsBar.tsx
+++ b/twake/frontend/src/app/scenes/Client/ChannelsBar/ChannelsBar.tsx
@@ -12,6 +12,9 @@ import './ChannelsBar.scss';
 import Shortcuts, { defaultShortcutsMap, ShortcutType } from 'app/services/ShortcutService';
 import ModalManager from 'app/components/Modal/ModalManager';
 import WorkspaceChannelList from './Modals/WorkspaceChannelList';
+import ScrollWithHiddenComponents from 'app/components/ScrollHiddenComponents/ScrollWithHiddenComponents';
+import HiddenNotificationsButton from 'app/components/ScrollHiddenComponents/HiddenNotificationsButton';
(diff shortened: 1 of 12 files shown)


twake/frontend/src/app/scenes/Client/ChannelsBar/ChannelsBar.tsx
```diff
@@ -37,17 +40,30 @@ export default () => {
...
-        <PerfectScrollbar component="div">
-          <ChannelsApps key={workspaceId} />
-          <Workspace key={'workspace_chans_' + workspaceId} />
-          <ChannelsUser key={companyId} />
-        </PerfectScrollbar>
-      )}
-      <Tutorial />
-      <Footer />
-    </Layout.Sider>
+    <ScrollWithHiddenComponents
```
RomaricMourgues, line 43: Here the scrollbar should not be for the whole bar but only the channels
Please keep the perfect scrollbar position and change the css of the go top go down buttons

RomaricMourgues (changes requested):

twake/frontend/src/app/services/languages/locale/en.js
```diff
@@ -1852,4 +1852,5 @@ export default {
   'scenes.client.mainview.tabs.tabstemplateeditor.title_tab_creation': 'Create a new tab',
   'scenes.client.mainview.tabs.tabstemplateeditor.title_tab_edition': 'Rename $1',
   'components.connectorslistmanager.add_connectors': 'Add connectors',
+  'components.notifications.hidden_notifications_button.children': 'Hidden notifications',
```
chamerling, line 1855: Such text message is not tech oriented, it should be "New messages" (or something like that)

stephanevieira75: Replaced to `New messages`, thank you @chamerling :smiley: 
Capture d’écran de 2020-12-15 16-55-49

PR #698, opened 2020-12-23
DRAFT: Working on tab's testing. Test with cassandra, almost done.

tissarni:
Test Create/Update with condition with/without ID seems don't work. Can't test Get all tabs cause of the first problem.

Co-authored-by: Christophe Hamerling <christophe.hamerling@gmail.com>

### DIFF
--- a/twake/backend/node/src/services/channels/services/tab/service.ts
+++ b/twake/backend/node/src/services/channels/services/tab/service.ts
@@ -11,12 +11,12 @@ import {
 } from "../../../../core/platform/framework/api/crud-service";
 import {  MemberService, TabService } from "../../provider";
 import { ChannelMember, ChannelTab, ChannelTabPrimaryKey } from "../../entities";
-import { ChannelExecutionContext, ChannelVisibility } from "../../types";
+import { ChannelExecutionContext, ChannelMemberType, ChannelVisibility } from "../../types";
 import { Channel } from "../../../types";
 import { DatabaseServiceAPI } from "../../../../core/platform/services/database/api";
 import { ResourcePath } from "../../../../core/platform/services/realtime/types";
 import Repository from "../../../../core/platform/services/database/services/orm/repository/repository";
-
+import {isDirectChannel} from "../../utils"
 
 export class Service implements TabService {
   version: "1";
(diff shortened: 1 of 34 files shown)


twake/backend/node/src/services/channels/services/tab/service.ts
```diff
@@ -45,6 +47,22 @@ export class Service implements TabService {
...
+
+    const userMember = await this.members.get({
+      channel_id: pk.channel_id,
+      company_id: pk.company_id,
+      workspace_id: pk.workspace_id,
+      user_id: context.user.id,
+    });
+
+    if (!userMember) {
+      throw CrudExeption.badRequest("User is not channel member");
```
chamerling, line 63: Replace with "User is now allowed to create tab"

twake/backend/node/src/services/channels/services/tab/service.ts
```diff
@@ -121,6 +151,17 @@ export class Service implements TabService {
   onDeleted(channel: Channel, tab: ChannelTab): void {
     console.log("Tab deleted", tab);
   }
+
+  async userMember(pk: ChannelTabPrimaryKey, context: ChannelExecutionContext): Promise<ChannelMember>{
```
chamerling, line 155: If you want to have a function like this, you can, but you have to name it correctly: `userIsChannelMember`, and you can return the result directly like:

```
async userIsChannelMember(user: string, channel: Channel): Promise<boolean>{
  return !!(await this.members.get({ CHANNEL }));
}
```

twake/backend/node/test/http/tabs.http
```diff
@@ -0,0 +1,90 @@
...
+### Delete a tab
+@deleteId = {{createTab.response.body.resource.id}}
+
+DELETE {{tabsURL}}/{{channelId}}/tabs/{{deleteId}}
+Authorization: Bearer {{authToken}}
+
+
+
+
+
```
chamerling, line 90: TO be removed

chamerling:
How did you tested it? 
1. Does it work on mongo?
2. Does it work on scylladb?

tissarni:
> How did you tested it?
> 
> 1. Does it work on mongo?
> 2. Does it work on scylladb?

I tested it with Mongo and Scylladb both started manually.

RomaricMourgues:
Replaced by #789

Issue #693, opened 2020-12-17
Need to regenerate default channels in brand new workspace


RomaricMourgues:
When we create a new workspace, there is not channel in it, we need to generate initial channels

PR #706, opened 2021-01-04
#693 Generate default channel in brand new workspace

stephanevieira75:
# Preview
Twake (8)


RomaricMourgues:
Thanks Stéphane! I’m thinking, this is not going to work with Twake Console because the workspace is going to be created from Backend in this case.
Let’s do a call this afternoon on this topic it can be an opportunity to do some Backend work

PR #711, opened 2021-01-06
Implement channel activity bold

RomaricMourgues:


twake/backend/node/src/services/channels/index.ts
```diff
@@ -19,8 +20,10 @@ export default class ChannelService extends TwakeService<ChannelServiceAPI> {
   public async doInit(): Promise<this> {
     const fastify = this.context.getProvider<WebServerAPI>("webserver").getServer();
     const database = this.context.getProvider<DatabaseServiceAPI>("database");
+    const pubsub = this.context.getProvider<PubsubServiceAPI>("pubsub");
+    console.log("PUBSUB FROM CHANNEL", pubsub);
```
RomaricMourgues, line 25: @chamerling this returns undefined, any idea ?
If I add a timeout of 5s it is not undefined, it looks like the service is not loaded yet at this time.
Working fine on the NotifiationServiceAPI

PR #756, opened 2021-01-12
📦 Data migration from 2020Q3 to 2020Q4

arkyek:
⚠️ Once merged this code changes the column type of channels:
connectors => text
members => text

ezgif com-gif-maker


chamerling:
1. Shouldn't we create a CLI for migrations? Pretty sure that it will not be the only one.
2. I suggest to use docker so that it can be tested and used everywhere (I do not have to install PHP and all the required dependencies). We already have a PHP image, maybe it can be used.

RomaricMourgues:
3. The migration should also be idempotent, running both time the migration should not duplicate data

RkAiman:
Moved here https://github.com/TwakeApp/Twake/pull/839

Issue #763, opened 2021-01-13
Implement Segment


RomaricMourgues:
Implement Segment in backend

- [x] Add generic parameters to track() and identify() use with other providers than Segment

- [x] Make sure the backend works if we do not provide segment or if key is invalid (in this case don't do anything when calling identify and track)

- [x] Get requests contexts in track and identify to get userId, companyId and other information from it

- [x] Implement tracker in the main features requested by Benoit

PR #803, opened 2021-01-20
#763 Implement Segment

stephanevieira75:
# Preview

https://user-images.githubusercontent.com/36481167/105204963-fe764680-5b44-11eb-97c6-ad2e9bd7e328.mp4

### DIFF
--- a/twake/backend/node/config/default.json
+++ b/twake/backend/node/config/default.json
@@ -9,10 +9,14 @@
     "database",
     "pubsub",
     "realtime",
-    "phpnode"
+    "phpnode",
+    "tracker"
   ],
   "logger": {
     "level": "debug"
+  },
+  "tracker": {
+    "segment": ""
   },
   "webserver": {
     "port": 3000,
--- a/twake/backend/node/package.json
+++ b/twake/backend/node/package.json
@@ -41,6 +41,7 @@
     "@babel/preset-env": "^7.11.5",
     "@babel/preset-typescript": "^7.10.4",
     "@types/amqp-connection-manager": "^2.0.10",
+    "@types/analytics-node": "^3.1.4",
     "@types/chai": "^4.2.12",
     "@types/config": "0.0.36",
     "@types/eslint": "^7.2.3",
@@ -76,7 +77,6 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
-    "@types/analytics-node": "^3.1.4",
     "amqp-connection-manager": "^3.2.1",
     "amqplib": "^0.6.0",
     "analytics-node": "^3.5.0",
--- a/twake/backend/node/package.json
+++ b/twake/backend/node/package.json
@@ -76,8 +76,10 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@types/analytics-node": "^3.1.4",
     "amqp-connection-manager": "^3.2.1",
     "amqplib": "^0.6.0",
+    "analytics-node": "^3.5.0",
     "cassandra-driver": "^4.6.0",
     "class-transformer": "^0.3.1",
     "config": "^3.3.2",
--- a/twake/backend/node/src/core/platform/framework/pubsub.ts
+++ b/twake/backend/node/src/core/platform/framework/pubsub.ts
@@ -1,0 +1,36 @@
+import { EventEmitter } from "events";
+import { TrackerEventActions } from "../../../core/platform/services/tracker/types";
+
+/**
+ * Event bus is used by the Pubsub decorators to publish data to be send to pubsub service.
+ * The pubsub service is the only subscriber of the event bus.
+ * By decoupling decorator and service, events will not be published when the service is not enabled.
+ */
+
+class TrackedEventBus extends EventEmitter {
+  constructor() {
+    super();
+  }
+
+  /**
+   * Subscribe is MUST only be used by the pubsub service
+   *
+   * @param listener
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  subscribe<T>(name: TrackerEventActions, listener: (data: T) => void): this {
+    return this.on(name, listener);
+  }
+
+  /**
+   * Publish must only be used by the Pubsub decorators
+   *
+   * @param data
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  publish<T>(name: TrackerEventActions, data: T): boolean {
+    return this.emit(name, data);
+  }
+}
+
+export const trackedEventBus = new TrackedEventBus();
--- a/twake/backend/node/src/core/platform/services/tracker/index.ts
+++ b/twake/backend/node/src/core/platform/services/tracker/index.ts
@@ -1,5 +1,12 @@
-import { Consumes, TwakeService } from "../../framework";
-import TrackerAPI, { IdentifyObjectType, TrackedEventType } from "./provider";
+import { Consumes, TwakeService, logger } from "../../framework";
+import TrackerAPI from "./provider";
+import { trackedEventBus } from "../../framework/pubsub";
+import {
+  TrackerEventActions,
+  TrackerDataListener,
+  IdentifyObjectType,
+  TrackedEventType,
+} from "./types";
 import Analytics from "analytics-node";
 
 @Consumes([])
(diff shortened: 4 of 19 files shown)


twake/backend/node/package.json
```diff
@@ -76,8 +76,10 @@
     "typescript": "^4.0.3"
   },
   "dependencies": {
+    "@types/analytics-node": "^3.1.4",
```
chamerling, line 79: This goes in `devDependencies`

twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -0,0 +1,39 @@
...
+    if ((await this.getAnalytics("segment")) && identity)
+      return (await this.getAnalytics("segment")).identify(identity, callback);
+  }
+
+  public async track(
+    tracker: TrackedEventType,
+    callback?: (err: Error) => void,
+  ): Promise<Analytics> {
+    if ((await this.getAnalytics("segment")) && tracker)
+      return (await this.getAnalytics("segment")).track(tracker, callback);
```
chamerling, line 24: You await the same thing too many times, in this function and above. Please refactor

twake/backend/node/src/core/platform/services/tracker/provider.ts
```diff
@@ -0,0 +1,30 @@
+import Analytics from "analytics-node";
+import { TwakeServiceProvider } from "../../framework";
+
+export type IdentityType =
+  | { userId: string | number }
```
chamerling, line 5: Do you need the first `|` ?

chamerling (changes requested):
Adding the tracker everywhere we need to track pollutes the code. I suggest another approach: Use a local pubsub component.

The local pubsub component is 'local' to the backend instance. It allows to publish and subscribe to topics like we do for the AMQP one:

- On each event that we think important enough, we publish it
- When activated, the tracker component subscribe to all the events which are useful to it, and then push them to segment or elsewhere.

By doing this, we do not pollute code with tracker, we are also able to add more subscribers which are not tracker ones but are interested in the events.

This is quite simple to implement, check what I did for distributed pubsub, you can do the same (cf twake/backend/node/src/core/platform/services/pubsub/bus.ts) but create this dedicated local event bus in a new component, it can be a single instance in twake/backend/node/src/core/platform/framework/bus.ts

twake/backend/node/src/core/platform/services/tracker/provider.ts
RomaricMourgues: This one is probably code formater, if yes you can let it like this @stephanevieira75

RomaricMourgues:
> Adding the tracker everywhere we need to track pollutes the code. I suggest another approach: Use a local pubsub component.
> 
> The local pubsub component is 'local' to the backend instance. It allows to publish and subscribe to topics like we do for the AMQP one:
> 
> * On each event that we think important enough, we publish it
> * When activated, the tracker component subscribe to all the events which are useful to it, and then push them to segment or elsewhere.
> 
> By doing this, we do not pollute code with tracker, we are also able to add more subscribers which are not tracker ones but are interested in the events.
> 
> This is quite simple to implement, check what I did for distributed pubsub, you can do the same (cf twake/backend/node/src/core/platform/services/pubsub/bus.ts) but create this dedicated local event bus in a new component, it can be a single instance in twake/backend/node/src/core/platform/framework/bus.ts

Just to be sure, this doesn't all rabbitmq ? I don't want to add rabbitmq calls for the tracker :)

chamerling:
@RomaricMourgues no it is not rabbit. What I call 'local' pubsub is an EventEmitter instance at the end

RomaricMourgues:
> @RomaricMourgues no it is not rabbit. What I call 'local' pubsub is an EventEmitter instance at the end

Perfect then !

twake/backend/node/package.json
stephanevieira75: Done :)

twake/backend/node/src/core/platform/services/tracker/index.ts
stephanevieira75: done 👍🏻

twake/backend/node/src/core/platform/services/tracker/provider.ts
stephanevieira75: Yes, it is code formater !

twake/backend/node/config/default.json
```diff
@@ -9,11 +9,15 @@
...
-    "phpnode"
+    "phpnode",
+    "tracker"
   ],
   "logger": {
     "level": "debug"
   },
+  "tracker": {
+    "segment": ""
+  },
```
chamerling, line 20: If we want multiple trackers (or to follow the same format as others), it should be:

```js
"tracker":
  "type": "segment",
  "segment": {
    "secret": "the secret"
  }
```

twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    return this;
+  }
+
+  private getVisibilityObject(visibility: string) {
+    return {
+      is_direct: visibility === "direct" ? true : false,
+      is_private: visibility === "private" ? true : false,
+      is_public: visibility === "public" ? true : false,
+    };
```
chamerling, line 85: Simple way to do:

```js
is_direct: !!(visibility === "direct)
```

twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -0,0 +1,118 @@
...
+      this.identify({
+        userId: data.user.id,
+      });
+      this.track(
+        {
+          userId: data.user.id,
+          event: TrackerEventActions.TWAKE_OPEN_CLIENT,
+        },
+        (err: Error) => (err ? logger.error("Error while tracking -->", err) : false),
+      );
```
chamerling, line 29: The `identity` and `track` functions are async, do you want to call identity then track or both in any order?

twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (analytics && tracker) return analytics.track(tracker, callback);
+  }
+
+  private async getAnalytics(trackerName: string): Promise<Analytics> {
+    const trackerKey = this.configuration.get<string>(trackerName, "");
+
+    if (!this.analytics && trackerKey) {
+      this.analytics = new Analytics(trackerKey);
+    }
+    return this.analytics;
```
chamerling, line 112: You should add a comment in the function saying that it may return undefined. Also, you must change the function return type because this.analytics can be undefined

twake/backend/node/src/services/channels/index.ts
```diff
@@ -7,7 +7,7 @@ import { DatabaseServiceAPI } from "../../core/platform/services/database/api";
 import { PubsubServiceAPI } from "../../core/platform/services/pubsub/api";
 
 @Prefix("/internal/services/channels/v1")
-@Consumes(["webserver", "database", "pubsub"])
+@Consumes(["webserver", "database", "pubsub", "tracker"])
```
chamerling, line 10: As we said, this should not be required here

twake/backend/node/src/services/notifications/index.ts
```diff
@@ -7,7 +7,7 @@ import { DatabaseServiceAPI } from "../../core/platform/services/database/api";
 import { PubsubServiceAPI } from "../../core/platform/services/pubsub/api";
 
 @Prefix("/internal/services/notifications/v1")
-@Consumes(["webserver", "database", "pubsub"])
+@Consumes(["webserver", "database", "pubsub", "tracker"])
```
chamerling, line 10: Also should not be required

twake/backend/node/src/services/notifications/services/engine/processors/new-channel-message/index.ts
```diff
@@ -9,6 +9,11 @@ import {
 } from "../../../../entities";
 import { ChannelMemberNotificationLevel } from "../../../../../channels/types";
 import { MentionNotification } from "../../../../types";
+import { trackedEventBus } from "../../../../../../core/platform/framework/pubsub";
```
chamerling, line 12: `trackedEventBus` is not a good name. Name it like `localBus` (same for other files)

twake/backend/node/src/core/platform/framework/pubsub.ts
```diff
@@ -0,0 +1,36 @@
+import { EventEmitter } from "events";
+import { TrackerEventActions } from "../../../core/platform/services/tracker/types";
```
chamerling, line 2: This is not related to tracker, this event bus is independant. The type should also be local to the pubsub.ts file (put it in this file)

twake/backend/node/src/core/platform/framework/pubsub.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor() {
+    super();
+  }
+
+  /**
+   * Subscribe is MUST only be used by the pubsub service
+   *
+   * @param listener
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
chamerling, line 20: I do not see any `any` :)

twake/backend/node/src/core/platform/framework/pubsub.ts
```diff
@@ -0,0 +1,36 @@
...
+  subscribe<T>(name: TrackerEventActions, listener: (data: T) => void): this {
+    return this.on(name, listener);
+  }
+
+  /**
+   * Publish must only be used by the Pubsub decorators
+   *
+   * @param data
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
chamerling, line 30: Same here, no `any`, so remove this comment

twake/backend/node/src/core/platform/framework/pubsub.ts
```diff
@@ -0,0 +1,36 @@
...
+  publish<T>(name: TrackerEventActions, data: T): boolean {
+    return this.emit(name, data);
+  }
+}
+
+export const trackedEventBus = new TrackedEventBus();
```
chamerling, line 36: `export const localEventBus = new EventBus();`

(and rename the class to `EventBus`)

PR #806, opened 2021-01-21
Add frontend logger and fix collection memory leaks

chamerling:


twake/frontend/src/app/scenes/Apps/Messages/Message/Parts/MessageContent.tsx
```diff
@@ -24,12 +23,16 @@ export default (props: Props) => {
...
+  const listener = Collections.get('messages').useListener(useState, [
     props.message?.id,
     props.message?.front_id,
     'msgcontent',
   ]);
 
+  useEffect(() => () => {
+    Collections.get('messages').removeListener(listener);
+  }, []);
+
```
RomaricMourgues, line 35: Good spot but I think we can improve this and fix the memory leak in a single place:
in src/app/services/Depreciated/observable.js line 17 (useListener method)
You can put the useEffect here, useState should not be passed as a variable either (I know this is done on a lot of places in the code with old collections but it's wrong)

chamerling: Oh YES! Let's try this

Issue #330, opened 2020-10-01
[Story] Change the group of a channel


BenoitTallandier:
Epic 

**As an internal user, I can move a channel into a group. When I do so, an activity message is sent in the channel.**

PR #823, opened 2021-01-27
📁  #330 Allow changing or create the group of a channel (#823)

stephanevieira75:
# Preview 
https://user-images.githubusercontent.com/36481167/106031280-62fa4e00-60cf-11eb-8714-32be6bb8fe8c.mp4

### DIFF
--- a/twake/frontend/src/app/scenes/Client/ChannelsBar/Modals/ChannelTemplateEditor.tsx
+++ b/twake/frontend/src/app/scenes/Client/ChannelsBar/Modals/ChannelTemplateEditor.tsx
@@ -1,8 +1,11 @@
 import React, { FC, useState, useEffect } from 'react';
 import Languages from 'services/languages/languages.js';
 import InputWithIcon from 'components/Inputs/InputWithIcon.js';
-import { ChannelType } from 'app/models/Channel';
-import { Select, Typography, Divider, Checkbox, Input } from 'antd';
+import { ChannelResource, ChannelType } from 'app/models/Channel';
+import { Select, Typography, Divider, Checkbox, Input, Button, Row, Col } from 'antd';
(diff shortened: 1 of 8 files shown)


chamerling:
Having an icon for group (the `+ folder` one) is nice. Anyway, the icon is not the right one, please find some which is more relevant :)

twake/frontend/src/app/scenes/Client/ChannelsBar/Modals/ChannelTemplateEditor.tsx
```diff
@@ -25,17 +28,34 @@ const ChannelTemplateEditor: FC<Props> = ({
...
+      channel_group: group,
       default: defaultChannel,
     });
   });
 
+  const getGroups = (channels: ChannelResource[]) => {
+    const groupsNames: string[] = [];
+    channels
+      .sort((a, b) => (a.data.channel_group || '').localeCompare(b.data.channel_group || ''))
+      .map((channel: ChannelResource) => {
```
chamerling, line 52: `map` is for transforming the input to output. Here you can use a `forEach` or use `map` like:

```js
const grousNames: string[] = channels.sort(sortFn).map(mapFn).filter(Boolean);
```

chamerling (changes requested):
Almost perfect 🤘🏻

twake/frontend/src/app/scenes/Client/ChannelsBar/Modals/ChannelTemplateEditor.tsx
stephanevieira75: Updated 👍🏻

stephanevieira75:
Here is the final design, i update the folder icon to a list icon and now show the divider between the icon and the channel name input. 😄 

Capture d’écran de 2021-01-28 11-37-48

PR #839, opened 2021-01-28
 🛫🛬 Migrate data from Q3 to Q4 as we changed our backend database

RkAiman:


twake/backend/node/src/services/channels/services/channel/cassandra.ts
```diff
@@ -87,8 +87,8 @@ export class CassandraChannelService implements ChannelService {
           name text,
           owner uuid,
           visibility text,
-          members frozen<set<text>>,
-          connectors frozen<set<text>>,
+          members text,
+          connectors text,
```
chamerling, line 91: Why? 😭

PR #896, opened 2021-02-09
Durable queues

chamerling:


RomaricMourgues:
@chamerling do we automatically ack in all the consumers we already had ?

chamerling:
@RomaricMourgues now yes

PR #899, opened 2021-02-09
🛠 More fixes

RomaricMourgues:


chamerling (approved):

chamerling:
Yes, but we are not speaking about the same tooltip, I am speaking about this one https://ant.design/components/tooltip/.
It allows to display the username when we put the mouse over the avatar, I guess we do not have to put the complete user card on this place, but at least displaying the username is nice enough.

I can rebase and push a PR, or you can grab the code from here: https://github.com/chamerling/Twake/commit/1aac08ea302bbef4db9580fa3214fcd999dac020

Just tell me

RomaricMourgues:
Oh yes, thanks I didn't saw that we missed this feature, I'll add it

PR #953, opened 2021-02-24
Create Node Users Service

chamerling:
It also introduces some stuff on companies and prepare the work for twake-console export.

### DIFF
--- a/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
+++ b/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
@@ -49,13 +49,16 @@ export function buildSelectQuery<Entity>(
     })
     .filter(Boolean);
 
-  const query = `SELECT * FROM ${options.keyspace}.${entityDefinition.name} WHERE ${[
+  let query = `SELECT * FROM ${options.keyspace}.${entityDefinition.name} WHERE ${[
     ...where,
     ...(buildComparison(findOptions) || []),
     ...(buildIn(findOptions) || []),
     ...(buildLike(findOptions) || []),
-  ].join(" AND ")}`;
+  ].join(" AND ")}`.trimEnd();
 
+  if (query.endsWith(" WHERE")) {
+    query = query.slice(0, -" WHERE".length)
+  }
   return query;
 }
 
--- a/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
+++ b/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
@@ -49,17 +49,14 @@ export function buildSelectQuery<Entity>(
     })
     .filter(Boolean);
 
-  let query = `SELECT * FROM ${options.keyspace}.${entityDefinition.name} WHERE ${[
+  const whereClause = `${[
     ...where,
     ...(buildComparison(findOptions) || []),
     ...(buildIn(findOptions) || []),
     ...(buildLike(findOptions) || []),
   ].join(" AND ")}`.trimEnd();
 
-  if (query.endsWith(" WHERE")) {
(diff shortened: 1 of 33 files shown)


twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
```diff
@@ -49,13 +49,16 @@ export function buildSelectQuery<Entity>(
...
     ...(buildComparison(findOptions) || []),
     ...(buildIn(findOptions) || []),
     ...(buildLike(findOptions) || []),
-  ].join(" AND ")}`;
+  ].join(" AND ")}`.trimEnd();
 
+  if (query.endsWith(" WHERE")) {
+    query = query.slice(0, -" WHERE".length)
+  }
   return query;
```
RomaricMourgues, line 62: Why don't you create a variable where before and then decide to append it or not ?

```
const where = [ ...where,
    ...(buildComparison(findOptions) || []),	    ...(buildComparison(findOptions) || []),
    ...(buildIn(findOptions) || []),	    ...(buildIn(findOptions) || []),
    ...(buildLike(findOptions) || []),	    ...(buildLike(findOptions) || []),
  ].join(" AND ")}`;

const query = "SELECT * FROM ${options.keyspace}.${entityDefinition.name} " + (where.length ? (" WHERE "+where) : "")
```

twake/backend/node/src/services/user/entities/company_user.ts
```diff
@@ -0,0 +1,48 @@
...
+// backward compatibility with PHP where companies used to be `group`
+export const TYPE = "group_user";
+
+/**
+ * Link between a company and a user
+ */
+@Entity(TYPE, {
+  primaryKey: [["group_id"], ["user_id", "id"]],
+  type: TYPE,
+})
```
RomaricMourgues, line 12: Here we'll have an issue when we'll want to get all companies of one user. For now it is okay, but we'll need to create an other table or store companies id directly in user entity.

twake/backend/node/src/services/user/web/schemas.ts
```diff
@@ -1,20 +1,28 @@
...
+    id: { type: "string" },
+    connected: { type: "boolean" },
+    creationdate: { type: "string" },
+    firstname: { type: "string" },
+    lastname: { type: "string" },
+    usernamecanonical: { type: "string" },
+    emailcanonical: { type: "string" },
+    phone: { type: "string" },
+    timezone: { type: "string" },
+    groups: { type: "array" },
```
RomaricMourgues, line 13: Here we should use a better schema for the future, I will document this.
At least we could replace groups by companies ?

RomaricMourgues (changes requested):
Small changes if you want, not necessary, rest seems good to me 👍

twake/backend/node/src/services/user/entities/company_user.ts
RomaricMourgues: And by the way: `[["group_id"], ["user_id", "id"]],` => `[["group_id"], "user_id", "id"],`

twake/backend/node/src/services/user/web/schemas.ts
chamerling: Yes and no :)
I just want to map the current schema. When import will be OK and when we will start using the node backend to manage companies, we will have a new, cleaner schema, and cleaner fields, and cleaner parameter names

twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
chamerling: Why not :)

twake/backend/node/src/services/user/entities/company_user.ts
chamerling: Yes!

RomaricMourgues (approved):

PR #977, opened 2021-03-02
CLI for console merge

chamerling:
Missing parts:

- [x] Create the Twake console HTTP client and integrate it in the command
- [x] Test with real console


RomaricMourgues:
@chamerling we didn't talk about how to do the link between Twake users and Console users, the table for this is "external_user_repository" and for companies "external_group_repository", it contains the console id and the Twake id.

chamerling:
@RomaricMourgues OK so for the external_* entities, there is a `service_id: string` to fill for both. I will set the value configurable from the CLI, anyway, what is the value we are currently using for the console `twake_console`?

RomaricMourgues:
It is "console" :)

RomaricMourgues (approved):
I read some part of it, it seems good to me good work

RomaricMourgues (approved):

PR #1015, opened 2021-03-11
🛠 Fix ctrl+K issues

RomaricMourgues:


twake/frontend/src/app/scenes/Client/ChannelsBar/Modals/WorkspaceChannelList.tsx
```diff
@@ -40,8 +40,7 @@ export default () => {
 
     if (event.key === 'Enter' && cursor >= 0) {
       const element = listService.list[cursor];
-
```
RomaricMourgues, line 43: Here there was an error when pressing enter and list was empty

twake/frontend/src/app/services/channels/channels.js
```diff
@@ -50,12 +50,13 @@ class Channels extends Observable {
   readChannelIfNeeded(channel) {}
 
   async openDiscussion(membersIds, companyId = null) {
+    membersIds = membersIds.map(e => e); //Copy original list
```
RomaricMourgues, line 53: And the issue on ctrl+K was here, because you pass directly the channel members list in "handleElementType" in WorkspaceChannelList.tsx 
So when we do a membersIds.push(UsersService.getCurrentUserId()); we do it on the original members list from the original  channel resource.

PR #1031, opened 2021-03-16
Default channels backend

chamerling:
This PR replaces https://github.com/linagora/Twake/pull/859

- [x]  E2E tests
- [x] PHP to push event in Rabbit when an user is added to a workspace

Event Payload:

```ts
type NewUserInWorkspaceNotification = {
  user: string;
  company_id: string;
  workspace_id: string;
};
```

Event topic: `workspace:user:added`

@RomaricMourgues point me to the right file for the PHP hook to add, I will implement it


chamerling:
Still needs to make some tests, there are more cases than expected but this is almost done :)

chamerling:
OK this is almost good. Just have to spot an "undefined" string coming from somewhere...
- User is well added to default channels when added to a workspace
- Default channels are displayed on all workspace users when they are created
- Channel is added to workspace user when its `is_default` flag is switched to `true`

chamerling:
TBD Missing for monday:

- [ ] Block default on private channels

Everything else is good 🍺

twake/backend/node/package.json
```diff
@@ -9,7 +9,7 @@
     "build:copy-assets": "npm run build:copy-config",
     "build:copy-config": "cp -R config dist",
     "dev": "tsc-watch --project . --noClear --outDir ./dist --onSuccess \"nodemon ./dist/server.js\" | pino-pretty",
-    "dev:cli": "tsc-watch --project . --noClear --outDir ./dist --onSuccess \"nodemon ./dist/cli/index.js console merge\" | pino-pretty",
+    "dev:cli": "tsc-watch --project . --noClear --outDir ./dist --onSuccess \"nodemon ./dist/cli/index.js workspace user\" | pino-pretty",
     "dev:cli-dry": "tsc-watch --project . --noClear --outDir ./dist --onSuccess \"nodemon ./dist/cli/index.js console merge --dry\" | pino-pretty",
     "dev:debug": "tsc-watch --project . --noClear --outDir ./dist --onSuccess \"nodemon --inspect=0.0.0.0 ./dist/server.js\" | pino-pretty",
```
RomaricMourgues, line 14: Do we keep this name dev:cli and dev:cli-dry ? Don't we want to have something like dev:cli:console:merge or even just `npm run dev:cli -- console merge` (and remove the "console merge" part hardcoded here) ?
Just a proposition

RomaricMourgues:
Can you add this business rule: 
- guests members are not added to default channels

You can get this information about guest or not guest in the company_user entity under the "is_externe" column.

And that said this column need to be updated, a "role" column with "admin" | "member" | "guest" would be better in the future.

twake/backend/node/package.json
chamerling: going to check this

chamerling: Well this is not possible since the cli is launched from tsc-watch which launches nodemon which is piped into pino-pretty...

chamerling:
While #1045 is not done, I added this commit to manage current isExternal flag (https://github.com/linagora/Twake/pull/1031/commits/4f41a69dcf95764212b30bd4a9f7a4b61ee337a9)

Then we will be able to update the default workflow.

Issue #360, opened 2020-10-06
[Story] Back / Next in browser


BenoitTallandier:
EPIC #308

**As a user, I can user back & next of my browser**

As a user, if I am in the channel A, and go to the channel B, if I click on "back" of my browser, I'm redirected to the channel A.

As a user, if I open parameters, and then I click on "back", I'm redirected where I was before opening parameters

tlle14:
Hello @RomaricMourgues @BenoitTallandier 
I don't see you describe the `Next button` behavior, is it ready to be tested? 
When I am on the previous screen, the `Next button` isn't enabled.

tlle14:
Related issue: 
https://github.com/TwakeApp/Twake/issues/820
https://github.com/TwakeApp/Twake/issues/821

hantt12:
Bug https://github.com/linagora/Twake/issues/1154

PR #1034, opened 2021-03-18
⬅️ #360 Fix Back / Next in browser

stephanevieira75:
# Preview:
https://user-images.githubusercontent.com/36481167/111649265-7d65c380-8804-11eb-8d32-02b3038e71f7.mp4

### DIFF
--- a/twake/frontend/src/app/services/RouterService.ts
+++ b/twake/frontend/src/app/services/RouterService.ts
@@ -257,6 +257,14 @@ class RouterServices extends Observable {
     }
     return false;
   }
+
+  pushTo(path: string, state?: unknown): void {
+    return this.history.push(path, state);
+  }
+
+  replace(path: string, state?: unknown): void {
+    return this.history.replace(path, state);
+  }
 }
 
 export default new RouterServices();
--- a/twake/frontend/src/app/services/RouterService.ts
+++ b/twake/frontend/src/app/services/RouterService.ts
@@ -258,7 +258,7 @@ class RouterServices extends Observable {
     return false;
   }
 
-  pushTo(path: string, state?: unknown): void {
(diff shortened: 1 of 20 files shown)


twake/frontend/src/app/services/RouterService.ts
```diff
@@ -257,6 +257,14 @@ class RouterServices extends Observable {
     }
     return false;
   }
+
+  pushTo(path: string, state?: unknown): void {
```
chamerling, line 261: This method name does not feel natural, why don't we keep `push`?

twake/frontend/src/app/services/login/login.js
```diff
@@ -321,8 +321,8 @@ class Login extends Observable {
         if (!no_reload) {
           Globals.window.location.reload();
         } else {
-          RouterServices.history.push(
-            RouterServices.pathnames.LOGIN + '?' + RouterServices.history.location.search.substr(1),
+          RouterServices.pushTo(
+            `${RouterServices.pathnames.LOGIN}?${RouterServices.history.location.search.substr(1)}`,
```
chamerling, line 325: We have to `substr` again and again and this looks like a bad API design

twake/frontend/src/app/services/RouterService.ts
stephanevieira75: Renamed to `push`

twake/frontend/src/app/services/login/login.js
RomaricMourgues: It is not necessary here, it should be:
```
`${RouterServices.pathnames.LOGIN}${RouterServices.history.location.search}`
```

stephanevieira75: updated 👍🏻

PR #1064, opened 2021-03-25
🧑‍⚖️ Linagora Developer's Certificate of Origin [READY TO BE MERGED]

shivam9593:

### DIFF
--- a/Contribute/Linagora Developer's Certificate of Origin
+++ b/Contribute/Linagora Developer's Certificate of Origin
@@ -18,16 +18,18 @@ To achieve this, we first need you to agree to this Developer Certificate of Ori
 
 By submitting a contribution to a LINAGORA project, you certify:
 
-a)   that this contribution is provided under the terms and conditions of the MIT license (https://opensource.org/licenses/MIT);
+a)   that this contribution is provided under the terms and conditions of the Creative Commons Zero (CC0) license (https://creativecommons.org/publicdomain/zero/1.0/legalcode);
 
-b)   being the rightful copyright owner of the contribution, and/or being fully and lawfully authorized to provide the contribution under the MIT license on behalf of all entities which participated in the creation of the contribution (for example your employer);
+b)   being the rightful copyright owner of the contribution, and/or being fully and lawfully authorized to provide the contribution under the CC0 license on behalf of all entities which participated in the creation of the contribution (for example your employer);
 
-c)   if the contribution is based upon a previous work, that this previous work is covered by an open source license which enables you lawfully redistribute your contribution and this previous work both under the MIT license;
+c)   if the contribution is based upon a previous work, that this previous work is covered by an open source license which enables you to lawfully redistribute your contribution and this previous work both under the LINAGORA project license (Just reach out to us if you have any doubts, and we will help you);
 
-d)   if moral rights apply to your contribution, to the extent permitted by applicable law, that you will not try to assert your moral rights in a manner which would, directly or indirectly, interfere with the use of your contribution either by LINAGORA or by any project user;
+d)  if your contribution is a derivative of one of our projects under a different license than the CC0 license, LINAGORA grants you a special authorization to provide this contribution to LINAGORA under the CC0 license regardless of the project license;
+
+d)   if moral rights apply to your contribution, to the extent permitted by applicable law, that you will not try to assert your moral rights in a manner which would, directly or indirectly, interfere with the use of your contribution either by LINAGORA or by any project user (though, of course, you will always be credited as a contributor to the project);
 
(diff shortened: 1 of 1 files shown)


RomaricMourgues (approved):

rouazana:
If I understand it correctly, it means that contributions are welcomed if they are provided under the MIT license.

As contributions are generally derivative work from the software, Twake, which is under AGPL with additional terms, how can a contributor put the derivative work under MIT license?

Hopefully I'm missing a point.

chamerling:
I agree with @rouazana. There is something strange with this.

shivam9593:
@rouazana @chamerling Very good points and thank you for bringing it up before we merged the request. 
I have asked the legal team to re-check that clause and correct it. Will keep you posted. 
But after the initial discussion, that clause is a mistake and it has to be changed. We somehow missed to think deeply about it while preparing the DCO.

shivam9593:
We have added an exception in the DCO, which is as following:

rouazana:
Thank you for taking account of my comments!

> I hope this answers your concerns. If there are any additional questions, please let me know.

That means that contributions will be double licensed MIT + Linagora's license. How do you plan to figure this double licensing? Will you add a MIT notice on each file modified by contributors?

shivam9593:
@rouazana No, it is not a double licensing but a re-licensing made possible by the MIT License. The MIT License allows re-licensing, and the contributions - wherever necessary - will be re-licensed under the project license. If the contribution - a single file never touched by LINAGORA, then the file will remain under the MIT License. But the project as a whole will be under the project license. 

Hopefully, it sounds clear. 🤞

rouazana:
> @rouazana No, it is not a double licensing but a re-licensing made possible by the MIT License. The MIT License allows re-licensing, and the contributions - wherever necessary - will be re-licensed under the project license. If the contribution - a single file never touched by LINAGORA, then the file will remain under the MIT License. But the project as a whole will be under the project license.
> 
> Hopefully, it sounds clear. crossed_fingers

At first I also thought about that, but there is this part in the MIT license:
```
The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.
```

Where do you forecast to put this notice regarding contributions?

shivam9593:
Indeed the copyright notice stays in the software contribution and inside the software, but the whole of our project is under our AGPL license. So yes, in fact, the contribution must remain under MIT with the copyright notice. 
Since it is only INSIDE the code, we can go on exploiting the project under AGPL.

rouazana:
> Indeed the copyright notice stays in the software contribution and inside the software, but the whole of our project is under our AGPL license. So yes, in fact, the contribution must remain under MIT with the copyright notice.
> Since it is only INSIDE the code, we can go on exploiting the project under AGPL.

Hence my question: where do you put this notice exactly?

There are basically 2 commons format for code files.
The format where you put the license at the beginning of each code file:
```
/**
... license ...
*/

code 

code

code

code
```

and the format where the license is implicit because declared at the root of the project (less recommended, but still pretty common)
```
code 

code

code

code
```

Suppose you have a contribution which is adding let's say 3 lines of code on 5 files. Where do you put the MIT notice?

shivam9593:
The first format is always the recommended one. To put the notice at the beginning of the code.

rouazana:
> The first format is always the recommended one. To put the notice at the beginning of the code.

Yes but if the MIT notice is at the beginning of the file, this means that all the file is now under MIT license.

In the recommended example:
```
/**
AGPL license with additional terms
(c) 2021 contributors
*/

code

code 

code

code
```

If there is a contribution, should it be:
```
/**
AGPL license with additional terms
(c) 2021 contributors
*/

code

code 

/**
The following lines are under MIT license
*/
contribution
/**
End of MIT license lines
*/
code

code
```
Honestly I find it crazy, the more there will be contributions, the more the code would be full of these license lines.

But given the double-licensing solution you propose, I don't see any other way. At least as it is not clear to me it should be described more completely in the DCO.

shivam9593:
The idea behind introducing the DCO was to remove the legal frills attached with a CLA. We tried to make the best DCO that will adhere to Linagora's philosophy and open-source license. However, you are right this will make our code chunky with lots of licensing texts. To address this problem, the best solution seems to be to accept the contributions under the terms and conditions of the Creative Commons Zero (CC0) license. (https://creativecommons.org/publicdomain/zero/1.0/legalcode) 

CC0 enables us to reuse the contributions without any need for notices or such (but contributions will still be acknowledged as such, because that's the right thing to do).

rouazana:
Thanks you for your work on this, I really appreciate that we move from CLA to DCO. I just want the process to be applicable in practice.

shivam9593:
Thanks to you, we were able to fix these potential issues before rolling them out to the open-source community.

PR #1067, opened 2021-03-25
🛤 Tracker update

RomaricMourgues:
If console is used, we should use Console Identifier for main tracking tools id as requested by marketing teams.

### DIFF
--- a/twake/backend/core/src/Twake/Users/Services/User.php
+++ b/twake/backend/core/src/Twake/Users/Services/User.php
@@ -324,6 +324,7 @@ class User
             "iat" => intval(date("U")) - 60*10,
             "nbf" => intval(date("U")) - 60*10,
             "sub" => $user->getId(),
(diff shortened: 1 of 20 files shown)


twake/backend/core/src/Twake/Users/Services/User.php
```diff
@@ -335,7 +336,8 @@ public function generateJWT($user, $workspaces = []){
             "type" => "refresh",
             "iat" => intval(date("U")) - 60*10,
             "nbf" => intval(date("U")) - 60*10,
-            "sub" => $user->getId()
+            "sub" => $user->getId(),
+            "csl_sub" => $user->getIdentityProviderId(),
```
chamerling, line 340: How bad is this name? `csl` means console, right? Why not just naming it `provider`?

RomaricMourgues: `provider_id` then ?

Issue #1066, opened 2021-03-25
Workspace & Company logo disapeared


BenoitTallandier:
**Describe the bug**
We don't know anymore the company & workspace logo

**To Reproduce**
1. Open Twake
2. Find a workspace which should have a logo
3. It doesn't

PR #1068, opened 2021-03-25
🩹 Release 434 patch

RomaricMourgues:
- [x] Missing workspaces and companies logos #1066 
- [x] Can't close channel creation modal after creation #1069
- [x] Can't create workspace (user invitation to workspace don't continue to workspace creation)

### DIFF
--- a/twake/backend/core/src/Twake/Workspaces/Entity/Group.php
+++ b/twake/backend/core/src/Twake/Workspaces/Entity/Group.php
@@ -62,7 +62,7 @@ class Group extends SearchableObject
 
     /**
      * @ORM\ManyToOne(targetEntity="Twake\Upload\Entity\File")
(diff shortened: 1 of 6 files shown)


RomaricMourgues:
@chamerling if you could just review the part to add requester as channel member it would be right.
I think there is code duplication but I don't know if I can change your "add creator into channel" and replace it by my "add requester into channel".

chamerling:
@RomaricMourgues it depends on how this is called compared to what is done here https://github.com/linagora/Twake/pull/1068/files#diff-dadb17dbeba816d165fbe481c88cae32a4dc796b14dee15cad64227b52731e88L577. 
In the current case, we add the creator as member at channel creation time. Why does it differ from this?

RomaricMourgues:
I will merge this because it is a patch, and I open an other PR to discuss the code duplication :)

PR #1143, opened 2021-04-12
Added additional subfields to last_message field of the channel

bobs4462:
New sub fields were added to last_message field of the channel entity

## Description
last_message field currently contains 4 fields: sender (user_id), date, title and text (username: message text). But currently in mobile application we need a little different data:
1. We need username separately (not as part of the text sub field)
2. Text sub field should contain only message itself without sender's name

In order to avoid breaking mobile notifications which is triggered by the same event as for updating last_message for the channel, I only appended the required fields to RabbitMQ data.

## Related Issue
https://github.com/linagora/Twake-Mobile/issues/433

## Motivation and Context
New mobile application design requires the above mentioned new fields

## How Has This Been Tested?
It wasn't tested

## Types of changes
Addition of the new fields to channel entity's last_message column, adjusted update logic

### DIFF
--- a/twake/backend/core/src/Twake/Discussion/Services/MessageSystem.php
+++ b/twake/backend/core/src/Twake/Discussion/Services/MessageSystem.php
@@ -580,6 +580,7 @@ class MessageSystem
 
         $title = "";
         $text = $this->buildShortText($message);
+        $body = $text; // we need original body just in case text gets updated further on
         if ($channel->getData()["is_direct"]) {
             $title = $senderName . " in " . $channel->getData()["company_name"];
         }else{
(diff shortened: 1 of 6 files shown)


twake/backend/core/src/Twake/Discussion/Services/MessageSystem.php
```diff
@@ -580,6 +580,7 @@ public function sendToNode($channel, $message, $did_create){
 
         $title = "";
         $text = $this->buildShortText($message);
+        $body = $text; // we need original body just in case text gets updated further on
```
RomaricMourgues, line 583: Why do you want the full body ? It can comes to errors with rabbitmq (rabbit only accept small messages) and I don't see the need here ? :)

RomaricMourgues: Ah sorrt I just understood:)

RomaricMourgues (approved):

PR #1195, opened 2021-04-22
💬 Implement messages in node part 1

RomaricMourgues:
#### Users Bookmarks (collection):
- [x] Rest
- [x] Realtime
- [x] Rest tests
- [x] Realtime tests

#### Thread creation:
- [x] Rest
- [x] Rest tests

#### Messages in thread (collection):
- [ ] Rest
- [ ] Realtime
- [ ] Rest tests
- [ ] Realtime tests

#### Messages views:
- [ ] Rest
- [ ] Realtime
- [ ] Rest tests
- [ ] Realtime tests

#### Don't forget
- [ ] Verify user have access to the participants he want to share the thread with (users and channels).
- [ ] Verify user can post to a thread (is in one of the participant channels)
- [ ] Verify always that the user is in the company requested in the URL (and apply it to other routes and other places of Twake)


twake/backend/node/test/e2e/messages/messages.user-bookmarks.realtime.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+                },
+                getContext(platform),
+              );
+            });
+            socket.on("realtime:resource", event => {
+              expect(event.type).toEqual("user_message_bookmark");
+              expect(event.action).toEqual("deleted");
+              done();
+            });
+          })
```
chamerling, line 123: This should be defined before the call to `userBookmarks.delete`

RomaricMourgues: Done, I moved the .emit("join") at the end so it does join -> realtime:join:success -> call service -> realtime:resource

twake/backend/node/test/e2e/channels/direct-channels.spec.ts
```diff
@@ -12,7 +12,7 @@ import { ChannelUtils, get as getChannelUtils } from "./utils";
 import { DirectChannel } from "../../../src/services/channels/entities/direct-channel";
 import { ChannelSaveOptions } from "../../../src/services/channels/web/types";
 
-describe("The direct channels API", () => {
+describe.only("The direct channels API", () => {
```
chamerling, line 15: This will skip all other tests

twake/frontend/public/index.html
```diff
@@ -120,6 +120,13 @@
           window.electron.remote.BrowserWindow.getAllWindows()[0]
         ) {
           window.electron_config = window.electron.remote.BrowserWindow.getAllWindows()[0];
+
+          const sw = electron.screen.getPrimaryDisplay().size.width;
+          const sh = electron.screen.getPrimaryDisplay().size.height;
+          const w = Math.min(Math.ceil(sw * 0.8), 1792);
+          const h = Math.min(Math.min(Math.ceil(sh * 0.8), 1120), w);
+          electron.remote.getCurrentWindow().setBounds({ x: Math.ceil((sw - w) / 2), y: Math.ceil((sh - h) / 2), width: w, height: h });
+          electron.remote.getCurrentWindow().center();
```
chamerling, line 129: Do we need some electron fix in this PR?

twake/backend/node/test/e2e/channels/direct-channels.spec.ts
RomaricMourgues: It seems I forgot this one, thanks. (but actually it still do every other test files, the "only" is only working inside a single file)

PR #1206, opened 2021-04-26
Please TS

chamerling:
@stephanevieira75 I can not run front without this fix


chamerling:
Fixed by https://github.com/linagora/Twake/pull/1203

PR #1220, opened 2021-04-28
Multiple message list fixes

chamerling:
TODO: 

- [x] Still need to address the attachment issue
- [x] Check all the inner components margin/padding since they cause height computing issues


chamerling:
https://user-images.githubusercontent.com/264403/116701320-140ece00-a9c8-11eb-8647-fd31cdc46c31.mp4

PR #1288, opened 2021-05-12
📠 Rewrite the text editor to support rich text edition features

chamerling:
This is a dirty WIP, anyway things are quite easy to implement once you understand how everything works...
The video shows mentions, emojis and text formatting (using key bindings, did not have time to add the toolbar but will be easy :) ). We will also be able to render anything else than text when selecting suggestion in the dropdown, etc, etc.

Code will be really simple compared to what we currently have. Next steps will also be markdown support, export/import, but there are existing modules for this.

https://user-images.githubusercontent.com/264403/118020082-02c7a900-b35a-11eb-9f24-53af924a488f.mov

Will clean and continue this after the spring break


RomaricMourgues:
Greaaaat ! Do you think we can render user mention with the user avatar and full name like a tag ?
Have a great holidays in the meanwhile :)

chamerling:
Yes we can render using the component of our choice. I was doing it but did
not have time to finish.

Le mer. 12 mai 2021 à 21:18, Romaric Mourgues ***@***.***> a
écrit :

> Greaaaat ! Do you think we can render user mention with the user avatar
> and full name like a tag ?
> Have a great holidays in the meanwhile :)
>

chamerling:
This one starts to be so big, I fixed a lot of things, I need to remove the old code and it should be good then. Anyway it will need some QA, because there are probably things I did not think about. I will write a "Definition of Done" but the BA will have to check it.

Will do a video before monday but you can start to review the code.

chamerling:
This one is now ready!

Issue #1317, opened 2021-05-27
Move fcm push to platform service


RomaricMourgues:
**For QA:**

Test if push notification works on QA (we migrated this service to Node)

PR #1321, opened 2021-05-27
🛠 Move fcm service to dedicated platform push service #1317 #1320

RomaricMourgues:


twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  async push(
+    devices: string[],
+    notification: PushMessageNotification,
+    options: PushMessageOptions,
+  ): Promise<void> {
+    const firebaseEndpoint = this.configuration.endpoint;
+    const firebaseApiKey = this.configuration.key;
+
+    for (const deviceIdentifier of devices) {
```
chamerling, line 19: I do not know how FCM works, but here you are looping over devices and sending as many HTTP requests as there are devices. I see that the `pushMessages` contains an array of `registration_ids` where you put only one itel when you can probably put more than one.

twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
```diff
@@ -0,0 +1,47 @@
...
+        const response = await (
+          await fetch(firebaseEndpoint, {
+            headers: {
+              "Content-Type": "application/json",
+              Authorization: `key=${firebaseApiKey}`,
+            },
+            body: JSON.stringify(pushMessage),
+          })
+        ).json();
+        logger.info(`${this.name} - Reply from FCM: ${JSON.stringify(response)}`);
```
chamerling, line 41: Should be `debug` when printing JSON. Also, `json()` on fetch response can take time (printing is also something which costs)

twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
```diff
@@ -0,0 +1,47 @@
...
+import { PushConfiguration, PushMessageNotification, PushMessageOptions } from "../../types";
+import { PushConnector } from "../connector";
+
+export default class FcmPushConnector implements PushConnector {
+  name = "FcmPushConnector";
+
+  constructor(readonly configuration: PushConfiguration["fcm"]) {}
+
+  async push(
+    devices: string[],
```
chamerling, line 12: You should assert than devices is defined, just because at runtime it can, and it will break your loop.

chamerling (commented):
This looks nice to have a platform service like this :)

twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
RomaricMourgues: Oups, it's an error my bad, thanks

twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
RomaricMourgues: Done

twake/backend/node/src/core/platform/services/push/connectors/fcm/service.ts
RomaricMourgues: Ok, logged only if response was not 200 now

chamerling (approved):

Issue #1301, opened 2021-05-21
Messages in node, API compatibility with PHP


RomaricMourgues:
**For QA**:
We moved the messaging backend from PHP to Node. This is a big change!
- [x] test we can send message and see them in realtime
- [x] test pin / unpin
- [x] test reaction to message
- [x] test send response under thread
- [x] test send message in thread view
- [x] test send files
- [x] test edit message (with or without file in it)
- [x] test create a call
- [x] test create a gif
- [ ] test remove a message
- [x] test channel creation and initial message(s)

PR #1323, opened 2021-05-28
🛠  Messages to node #1301 bux fixing

RomaricMourgues:
- [x] Check applications still works with php api forwarded to node / ephemerals
- [x] Remove threads when empty
- [x] Remove message fix
- [x] Ephemerals: closable by all options fix
- [x] Thread update (number of answers etc) realtime update
- [x] Pin / reaction not working as expected bug
- [x] Convert maximum possible Twacode <-> Blocks
- [x] Init message order is wrong (appear after "you have been added to the channel") -> ~~must also be moved to node~~ it is depreciated and must be done by frontend from now on (so stored in frontend as a monkeyfix in the meanwhile)
- [x] Drag drop to other threads and out of thread too

### DIFF
--- a/twake/backend/node/src/services/messages/services/messages/utils.ts
+++ b/twake/backend/node/src/services/messages/services/messages/utils.ts
@@ -85,7 +85,10 @@ export function getDefaultMessageInstance(item: Partial<Message>, context: Threa
   });
 
   if (context.user.server_request) {
-    instance = _.assign(instance, item);
+    instance = _.assign(
+      instance,
+      _.pickBy(item, v => v !== undefined),
+    );
   }
 
   return instance;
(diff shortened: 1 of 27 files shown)


twake/backend/node/src/services/messages/services/messages/utils.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (selectedReactions.includes(key)) {
+      reactions[key].count++;
+      reactions[key].users.push(userId);
+    }
+    if (reactions[key].count === 0) {
+      delete reactions[key];
+    }
+  }
+  message.reactions = Object.values(reactions);
+}
```
RomaricMourgues, line 47: Probably better ways of doing this (@chamerling is going to 🔪  me because of not using map)

And yeah, the guy comments on its own PR

chamerling: AHHHHHHHHHHHHH!!!!!!

PR #1324, opened 2021-05-28
🛠  Localstorage fix

RomaricMourgues:
Improve the local storage clear method to only remove twake stored items.
In the future all twake local stored items must be prefixed with `twake:`.


chamerling (commented):
The prefix should be added by the LocalStorage service ( LocalStorage from 'services/localStorage.js) and we should add a rule to not use window.localstorage in all other files

RomaricMourgues:
> The prefix should be added by the LocalStorage service ( LocalStorage from 'services/localStorage.js) and we should add a rule to not use window.localstorage in all other files

Done

PR #1325, opened 2021-05-28
🏢 Tests and endpoint for public request /companies/:companyId

romanesko:
GET /companies/:company_id
Get a company by id and public information

```js
{
  "resource": {
    "id": string,
    "name": string,
    "logo": string,
    "plan": {
      "name": string,
      "limits": {
        "members": number,
        "guests": number,
        "storage": number
      }
    },
    "stats": {
      "created_at": timestamp,
      "total_members": number,
      "total_guests": number,
    }
  },
  "websockets": [] //Empty for now
}
```


twake/backend/node/src/services/user/web/routes.ts
```diff
@@ -50,8 +50,6 @@ const routes: FastifyPluginCallback<{
   fastify.route({
     method: "GET",
     url: "/companies/:id",
-    preHandler: accessControl,
-    preValidation: [fastify.authenticate],
```
chamerling, line 54: Why?

twake/backend/node/test/e2e/users/users.spec.ts
```diff
@@ -3,7 +3,8 @@ import { init, TestPlatform } from "../setup";
 import { TestUsers } from "./utils";
 
 describe("The /users API", () => {
-  const url = "/internal/services/users/v1/users";
+  jest.setTimeout(30000);
```
chamerling, line 6: Should be in JEST configuiration

twake/backend/node/test/e2e/users/users.spec.ts
```diff
@@ -278,23 +279,55 @@ describe("The /users API", () => {
...
         method: "GET",
-        url: `/companies/${companyId}`,
+        url: `${url}/companies/${companyId}`,
       });
       expect(response.statusCode).toBe(200);
+
+      const json = response.json();
+      // expect(json).toMatchObject({ resource: expect.any(Array) });
+
+      console.log(json.resource);
```
chamerling, line 304: Useless

twake/backend/node/test/e2e/users/users.spec.ts
romanesko: But we don't have any for the project. I added jest.config.js once, but Romaric said it could be the reason all tests failed  and I removed it

twake/backend/node/src/services/user/web/routes.ts
romanesko: from the task:
_Get a company by id and public information (this route is public and doesn't need to be authenticated)_

RomaricMourgues: Yes, companies general identity is public in Twake php. We can change this later of course (it is used for public shared links like folders or tabs).
We keep it public until we have a concept of "public JWT in-url token" or something.

twake/backend/node/test/e2e/users/users.spec.ts
RomaricMourgues: If you look at "test:e2e" in package.json you will see ` --testTimeout=20000`, just add it to your npm command

romanesko: yep, but the problem is that I do TDD and run tests directly from IDE in debug mode (so I don't use package.json scripts for this), and Idea creates configurations for every test run automatically. but that's ok, I can handle it on my side. I've removed this line/

chamerling (approved):
Looks good!

RomaricMourgues (changes requested):
I will merge as soon as:
- Tests pass (it seems there is an error with the user spec tests)
- You put back the package-lock.json (just do a npm install it will regenerate automatically)

Issue #1230, opened 2021-04-30
Migration script for messages 


stephanevieira75:
[ ] - Get php message
[ ] - Create migration folder
[ ] - Generate messages command
[ ] - Create entity

PR #1326, opened 2021-05-28
⏩ #1230 Migration script for messages 

stephanevieira75:


twake/backend/node/src/services/messages/api.ts
```diff
@@ -12,15 +12,21 @@ import {
...
   ThreadExecutionContext,
   MessageWithReplies,
 } from "./types";
+
 import { ParticipantObject, Thread, ThreadPrimaryKey } from "./entities/threads";
 import { Message, MessagePrimaryKey } from "./entities/messages";
+import {
+  PhpMessage,
+  PhpMessagePrimaryKey,
+} from "../../cli/cmds/migration_cmds/php-message/php-message-entity";
```
chamerling, line 29: This is something that MUST not happen. The runtime depends on some CLI code which is bad

chamerling: A solution will be to have a `legacy` folder where both CLI and the message API depends on

twake/backend/node/src/services/console/processing/merge.ts
```diff
@@ -54,7 +55,7 @@ export class MergeProcess {
   ) {
     this.client = new ConsoleHTTPClient(consoleClientParameters, dryRun);
   }
-
+  //
```
RomaricMourgues, line 58: Why modifying this code ?

twake/backend/node/src/services/messages/entities/messages.ts
```diff
@@ -88,7 +88,7 @@ export class Message {
 
 export type MessageReaction = { count: number; name: string; users: string[] };
 
-export type MessageOverride = { title?: string; picture?: string };
+export type MessageOverride = { title?: string; picture?: string; type?: string };
```
RomaricMourgues, line 91: What is type here?

PR #1359, opened 2021-06-17
🧹 End of spring cleanup before working on guests

chamerling:
WIP Nothing fixed for now, just cleaning code


chamerling:
OK so before fixing the guests issues, I guess that we can merge this HUGE cleaning (I will stop here for this one because we need to find limits :p )

RomaricMourgues (approved):
Seems good to me, if no imports are missings then I think it is good to go!

PR #1361, opened 2021-06-18
Globals refactoring

chamerling:
Move globals to TS, cleaning code around it and introducing decorator to stop doing registering services manually


RomaricMourgues (approved):
Looks good to me

PR #1363, opened 2021-06-18
🛠 Ability to create global indexes

RomaricMourgues:
This is used for the user table as a temporary solution. Reminder: the goal is to not use any advanced indexes on Scylla or Cassandra because Cassandra marked them as experimental, then depreciated.

### DIFF
--- a/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/cassandra.ts
+++ b/twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/cassandra.ts
@@ -217,7 +217,7 @@ export class CassandraConnector extends AbstractConnector<
 
     if (primaryKey.length === 0) {
       logger.error(
-        "services.database.orm.cassandra - Primary key was not defined for table " + entity.name,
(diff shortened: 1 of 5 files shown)


twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/cassandra.ts
```diff
@@ -289,6 +290,31 @@ export class CassandraConnector extends AbstractConnector<
       result = false;
     }
 
+    // --- Create indexes --- //
+    if (entity.options.globalIndexes) {
+      for (const globalIndex of entity.options.globalIndexes) {
+        const indexName = globalIndex.join("_");
+        const indexDbName = "index_" + md5(indexName);
```
chamerling, line 297: template literal :p

chamerling (approved):
Tiny comment and we are good!

twake/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/cassandra.ts
RomaricMourgues: Rhoo, pasted from php I missed this one :(

PR #1370, opened 2021-06-22
🌟 Feature/workspace pending users

romanesko:
pending users support (invite/retrieve/delete)
processing accepted users to workspaces (without endpoint yet)


RomaricMourgues:
You have two failed tests on mongodb, could you look at it ?

twake/backend/node/src/services/user/services/users/service.ts
```diff
@@ -80,15 +80,17 @@ export class UserService implements UsersServiceAPI {
     return this.repository.find(findFilter, findOptions);
   }
 
+  getByEmails(emails: string[]): Promise<User[]> {
+    return this.repository
+      .find({}, { $in: [["email_canonical", emails]] })
```
RomaricMourgues, line 85: Found it, the "IN" operator doesn't work with indexes. Could you fetch entities one by one instead of using a in ?

PR #1394, opened 2021-06-28
🧹 User AyncGet refactoring

chamerling:
This has to be discussed. On first step I wanted to extract parts where asyncGet is used and then move to recoiljs. Then I saw all the complexity of Collections and what it means to move from collections to recoil and all the work which it implies...


chamerling:
@RomaricMourgues @stephanevieira75 this one is ready

stephanevieira75 (approved):

PR #1408, opened 2021-06-30
🐰 Add RXJS local pubsub implementation

chamerling:
Implementation of pubsub using RXJS. The configuration has changed but defaults to AMQP implementation for now.

### DIFF
--- a/twake/backend/node/config/default.json
+++ b/twake/backend/node/config/default.json
@@ -83,9 +83,13 @@
     }
   },
   "pubsub": {
-    "urls": [
-      "amqp://guest:guest@localhost:5672"
-    ]
+    "// possible 'type' values are": "'ampq' or 'local'",
+    "type": "amqp",
+    "amqp": {
+      "urls": [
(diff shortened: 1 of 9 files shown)


RomaricMourgues:
Do you think the default configuration should be LOCAL ? For community purpose, or do we put the local as environment variable in docker-compose.yml.onpremise maybe

chamerling:
> Do you think the default configuration should be LOCAL ? For community purpose, or do we put the local as environment variable in docker-compose.yml.onpremise maybe

We can start to use it ourselves and then switch to default when we are sure that all is OK

twake/backend/node/config/default.json
```diff
@@ -83,9 +83,13 @@
     }
   },
   "pubsub": {
-    "urls": [
-      "amqp://guest:guest@localhost:5672"
-    ]
+    "// possible 'type' values are": "'ampq' or 'local'",
```
RomaricMourgues, line 86: So this is how you do comments in json!

RomaricMourgues (approved):

Issue #429, opened 2020-10-19
Fix linter warnings


stephanevieira75:
![Capture d’écran de 2020-10-19 13-37-32.png](https://images.zenhubusercontent.com/5f4e3d2d91dca0478164f0b5/b2791534-bec8-4e01-8a80-544df75c0357)

RomaricMourgues:
Fix only the ts files. No need to fix linter on .js files right now

PR #1421, opened 2021-07-05
#429 Fix useEffect array dependencies in Observable.ts

stephanevieira75:


RomaricMourgues:
That's a very important change here, it is probably not the issue origin. If you change this you could break everything on Twake (remove view refresh etc).
Could you add more details about where this fix comes from ?

stephanevieira75:
> is probably not the issue origin. If you change this you could break everything on Twake (remove view refresh etc).

I changed it without doing it on purpose here 92292c1a8ca32652709caea7f20859d3578cb7a1

Back to normal now !

PR #1459, opened 2021-07-16
📱 Device management endoints

romanesko:


RomaricMourgues:
@romanesko I propose we inverse (again) the primary key order. My idea is to put the devices in "cache" directly into the user entity and use the "device" entity only to avoid duplication.

- It means we use "id" as primary key for device entity.
- And we use user.devices = [string, string, ...] (user entity).

Each time someone want to register a device token, it remove it from the current user that use it, and then it set the new user id and add it to the users device tokens. This will prevent two users using the same device token (that happens a lot if logout wasn't properly done for instance). And make things faster when looking for user devices before sending notifications (as it is directly in user entity).

What do you think, was it clear ?

romanesko:
> What do you think, was it clear ?

Yep!

RomaricMourgues:
Started to review it, seems perfect to me :)

RomaricMourgues:
If not already done, could you add a get my devices endpoint for the user ? And update the notion doc accordingly, thanks!

RomaricMourgues:
You have an error in your tests because token is a reserved keyword. I need to see if we can fix it on the ORM or need to change the column name. It did work on php so probably some ORM changes to do. You'll tell me.

romanesko:
> could you add a get my devices endpoint for the user ? And update the notion doc accordingly, thanks!
done

> I need to see if we can fix it on the ORM or need to change the column name.
I changed the column name to `id` (as we do it in common for pk)

RomaricMourgues:
@romanesko `Cannot find module 'bcrypt' from 'src/utils/password-encoder.ts'"},"msg":"/usr/src/app/src/services/user can not be loaded"` could you check you pushed the package.json and package.lock correctly ?

RomaricMourgues:
You did not push the package.json file according to changed file in this PR

Issue #1476, opened 2021-07-22
Login / Token endpoint update


RomaricMourgues:
- I need to rename the current /token route to /login
- and to create a new /internal/services/console/v1/token endpoint that will manage the renewal of the Twake JWT from an access_token or a refresh_token in the headers.
- and to rename access_token to remote_access_token in the new /login endpoint to avoid misleading future developers.

PR #1479, opened 2021-07-23
🌟 #1476 Login / Token endpoint update 

romanesko:
renamed /token route to /login
new /internal/services/console/v1/token endpoint that  renewal of the Twake JWT from an access_token or a refresh_token in the headers.
renamed access_token to remote_access_token in the new /login endpoint to avoid misleading future developers.


RomaricMourgues:
@romanesko Did you try to import a frontend component in the backend code ? 😜
https://github.com/linagora/Twake/pull/1479/checks?check_run_id=3143899077#step:3:73

twake/backend/node/src/services/console/web/controller.ts
```diff
@@ -34,15 +32,33 @@ export class ConsoleController {
...
+    }
+
+    const token = authHeader.substr(7).trim();
+
+    let data: JWTObject = null;
+    try {
+      data = this.authService.verifyToken(token);
+    } catch (e) {
+      throw CrudExeption.forbidden(e.message);
+    }
```
chamerling, line 57: Checking token must be in the route preprocessing. There are places to do this with fastify, then it can be reused in other routes

twake/backend/node/src/services/console/web/routes.ts
```diff
@@ -55,12 +55,20 @@ const routes: FastifyPluginCallback<{
 
   fastify.route({
     method: "POST",
-    url: "/token",
+    url: "/login",
     // preHandler: [accessControl],
```
chamerling, line 59: This is the place to check token

PR #1489, opened 2021-07-27
set limit to 10000 if limit is not a number ("undefined")

romanesko:


twake/backend/node/src/services/channels/web/routes.ts
```diff
@@ -46,11 +46,22 @@ const routes: FastifyPluginCallback<{ service: ChannelServiceAPI }> = (
...
   };
 
+  const validateQuery = async (
+    request: FastifyRequest<{
+      Params: BaseChannelsParameters;
+      Querystring: PaginationQueryParameters;
+    }>,
+  ) => {
+    if (isNaN(+request.query.limit)) {
+      request.query.limit = "100000";
```
RomaricMourgues, line 56: Could we set the limit to "100" by default ?

PR #1348, opened 2021-06-11
🛠  Move search service

RomaricMourgues:
- [x] Move search service in a dedicated folder
- [x] Use localEventBus to send created / removed entities and tables
- [x] Implement search feature for mongo
- [x] Implement search feature for es

PR #1490, opened 2021-07-27
🛠 Fixes 2021.Q3.600

### LINKED ISSUE #1488: Can create the same direct channel indefinitively

RomaricMourgues:
- [x] #1488 
- [x] #1487 Fixed by #1348 (uuid/string format mismatch)
- [x] Move a message in other thread
- [x] Move message out of thread make it desapear
- [x] Fix message editor UI
- [x] Broken message editor counter
- [x] Colored background while loading

After
Screenshot 2021-07-28 at 15 19 09

Before
Screenshot 2021-07-28 at 15 17 50


stephanevieira75:
I like this ! But what happens when the message count appears ?

RomaricMourgues:
> I like this ! But what happens when the message count appears ?

Good spot, it is broken a bit :,(

PR #1532, opened 2021-08-04
Introduce recoil

### LINKED ISSUE #1537: Add a state manegement library
After some search, it seems that there is nothing really better than recoil. 
The PR https://github.com/linagora/Twake/pull/1532 introduces it.

chamerling:
Fixes #1537 

There are many things which are bad in the current codebase, but the main one is that the 'state' is updated in 'Collections' by adding some attributes to objects. recoil is not happy with that, just because atom values are immutable.


twake/frontend/src/app/state/recoil/atoms/CurrentCompany.ts
```diff
@@ -0,0 +1,16 @@
...
+import UserContextState from "../../UserContextState";
+
+const currentWorkspaceEffect: AtomEffect<CompanyType| undefined> = ({ onSet }) => {
+  onSet(company => UserContextState.company = company);
+};
+
+export const CurrentCompanyState = atom<CompanyType | undefined>({
+  key: 'CurrentCompanyState',
+  default: undefined,
+  effects_UNSTABLE: [
```
RomaricMourgues, line 13: After all this is just a question of is it more unstable than current Twake or less unstable :p

RomaricMourgues (approved):
I like what I see

twake/frontend/src/app/state/recoil/atoms/CurrentCompany.ts
chamerling: You get it 😸

stephanevieira75 (approved):

chamerling:
So far I wanted to go deeper and remove all the calls to Workspace and Group services which to replace them by calling the new context, but this is quite impossible until we refactor the `LoginService` which is full of dirty fixes...
I am stopping here for this one

RomaricMourgues:
🎉 🍻

PR #1563, opened 2021-08-13
WIP Refactoring the authentication process

chamerling:


RomaricMourgues:
@chamerling tell me if you want to update the QA at some point.
You can also do it yourself, just force-push to the develop branch from the feature/2021.Q3.dev.

chamerling:
I will have to do some tests with the console so yes updating the QA will be nice at some point. Still need some time for this refactoring, there are tons of things to change...
So, force push develop is enough? What is the delay to have it deployed on QA then?
Anyway, being able to run frontend on localhost and using everything else from QA will be nice

RomaricMourgues:
> So, force push develop is enough? What is the delay to have it deployed on QA then?

Once the ci ends, it takes 30 minutes but you can ping me to force it to update.

> Anyway, being able to run frontend on localhost and using everything else from QA will be nice

You can do this only with development server (because localhost must be whitelisted in the OIDC provider config).

twake/frontend/src/app/services/login/LoginService.ts
```diff
@@ -0,0 +1,88 @@
+import Api from '../Api';
+import { TwakeService } from '../Decorators/TwakeService';
+import Globals from '../Globals';
+import { Login as DeprecatedLogin } from './login';
+import AuthService from '../Auth/AuthService';
+import { AuthProvider } from '../Auth/provider/AuthProvider';
+
+type ExternalProviderType = 'openid' | 'cas' | string;
```
chamerling, line 8: @RomaricMourgues do we still want to support CAS and OPENID? If not I will remove all the related code

RomaricMourgues:
No we don’t, this should be supported by the Console directly not by us.

We only support “Console" and “Internal” login.

chamerling:
@RomaricMourgues I need to test. Can you tell me where is the machine which is updated when pushing to develop? (also, I need to get the OIDC config etc). Do we have this on Notion? (then please take real holidays ☀️)

RomaricMourgues:
Hello @chamerling sorry to disturb you in holidays, but I can't get the current status of this issue, last status was "waiting to be cleaned" according to @BenoitTallandier . Is it ready to merge? (If you don't answer then I will consider it as not ready and we wait for next monday)

RomaricMourgues:
In conclusion @chamerling is it breaking something or not ? If not do we merge it as it at-least refactored the code ?

chamerling:
@RomaricMourgues I need to rebase and I am still trying to test with the console. I am not far from something...

PR #1567, opened 2021-08-16
1474 workspace and company counters

### LINKED ISSUE #1474: Workspace and company counters
Currently number of users in a workspace or in a company is not up to date. We need to get it right.

- [x] Add a counter table (special Cassandra table type)
- [x] Increment / decrement the counter on user added / removed
- [x] Randomly (ex. 1 time our of 10 for instance) or when we list the users, decide to do a re-count (asynchronously preferably) and add the delta value to the counter to set it to the right value.

romanesko:
Resolves #1474


chamerling:
I am a bit disappointed on how we have to randomly update counters. What about having a CountService which does this:

- Update counter every N calls
- Update counter if last update is too old

Both N calls and update time can be held in memory.

I suggest something like this:

```js
const updater = countService.addCounterUpdater(handler, n, period);
```

where:
- `handler` is the function to call to update the counter.
- `n` is the number of calls to wait to update
- `period` is the max period to wait between updates

Then every time you need to ask for update, you can call `update()` which has been sent back by the `addCounterUpdater` function.

Note that we can just have a single countService for all.

Also, when you want to increase/decrease values, you do not have to call it like you do, just call the counter service.

WDYT @romanesko ?

romanesko:
I'm not happy with that realization, I did it the way Romaric asked me to. I commented it here:
https://app.zenhub.com/workspaces/twake-5f06c5bf7796310019354875/issues/linagora/twake/1474

chamerling:
What I suggested is to really decouple the code to update counters. Then one day we will be able to update this counter service to better fit. This is not ideal, I agree, but code will be cleaner

romanesko:
You are right. I'll redo this.

RomaricMourgues:
We need to have a dedicated service to manage counters correction and recount +1

The strategy of random recount is from Benoit from James team, they do random recount of unread emails, I can’t find the original ADR yet, but you can ask him for more details.

RomaricMourgues:
Small update on your proposal:

```typescript
//Make second argument a option object and also make it optional (with default values in the service)
const updater = countService.createUpdater(handler, {maxCalls: n, maxPeriod: period});

//To use the updater:
await updater.add(1);
await updater.add(-1);
```

Do I have you validation for this change @chamerling ?

twake/backend/node/src/services/workspaces/services/workspace/service.ts
```diff
@@ -63,6 +75,21 @@ export class WorkspaceService implements WorkspaceServiceAPI {
...
+
+    this.workspaceCounter = await this.counters.getCounter<WorkspaceCounterEntity>(
+      workspaceCountersRepository,
+    );
+
+    this.workspaceCounter.reviseCounter(async (pk: WorkspaceCounterPrimaryKey) => {
+      return this.workspaceUserRepository
+        .find({ workspace_id: pk.id })
+        .then(a => a.getEntities().length);
+    });
```
RomaricMourgues, line 91: This will not work if there is pagination (more than 100 members) could you add a loop over the pages like we did here ? https://github.com/stephanevieira75/Twake/blob/1bcd2ad9eebde1fb087be77a6b1398aed9adaba2/twake/backend/node/src/cli/cmds/migration_cmds/php-message/message-migrator-service.ts#L47

romanesko: 👀
sure!

romanesko:
moved to another PR: https://github.com/linagora/Twake/pull/1593

Issue #1539, opened 2021-08-05
The not-yet-login user doesn't appear in the workspace member list


tlle14:
# Precondition
- User is created on Console 
- User has not logged in yet

# Actual Result
On the default workspace member list, the new user doesn't appear.
If the user is added to the workspace, the console log screen said he has already been in there.
![image_2021_08_05T08_27_03_569Z](https://user-images.githubusercontent.com/69950655/128327843-1d2c2268-d5ff-4185-839c-b5d7b006d3b8.png)

# Expected Result
Even the user has not logged in yet, he should display in the member list of the workspace

# Environment
QA environment
Chrome Ubuntu

RomaricMourgues:
Note: we need to refresh after 2 seconds this list because it is async behind (waiting for refactor of this pge with node backend)

PR #1595, opened 2021-09-01
🌟 #1539 Refactor workspace users manager frontend

stephanevieira75:
https://user-images.githubusercontent.com/36481167/132562208-42c762b0-d88a-477b-b524-2aa18fe5eb84.mp4


twake/backend/node/src/services/workspaces/web/routes.ts
```diff
@@ -89,10 +89,10 @@ const routes: FastifyPluginCallback<{
       if (!workspace) {
         throw fastify.httpErrors.notFound(`Workspace ${workspaceId} not found`);
       } else {
-        throw fastify.httpErrors.forbidden("Only workspace admin can perform this action");
+        throw fastify.httpErrors.forbidden("You must be member of the workspace");
       }
-    } else if (workspaceUser.role !== "admin") {
-      throw fastify.httpErrors.forbidden("Only workspace admin can perform this action");
+    } else if (workspaceUser.role !== "moderator") {
```
RomaricMourgues, line 94: Could we replace this by the utils function "hasWorkspaceAdminLevel()" ?

twake/backend/node/src/services/workspaces/web/controllers/workspaces.ts
```diff
@@ -191,7 +191,7 @@ export class WorkspacesCrudController
 
     const workspaceUserRole = await this.getWorkspaceUserRole(request.params.id, context);
 
-    if (workspaceUserRole !== "admin") {
+    if (workspaceUserRole !== "moderator") {
```
RomaricMourgues, line 194: Could we replace this by the utils function "hasWorkspaceAdminLevel()" ?

RomaricMourgues (changes requested):
Also it seems we are missing the frontend side to switch from "admin" to "moderator".
+ I asked Roman to make sure every company admin is workspace moderator, I don't see this code anywhere yet

twake/backend/node/src/services/workspaces/web/controllers/workspaces.ts
stephanevieira75: done

twake/backend/node/src/services/workspaces/web/routes.ts
stephanevieira75: done

RomaricMourgues (approved):

PR #1637, opened 2021-09-20
📈 Fix anonymousId in tracker

RomaricMourgues:
Our users can enable anonymous tracking for us to keep track of the platform usage. But currently all ids are "anonymous" (so we cannot count unique events).
Now we will have md5(userId) as anonymous id.

### DIFF
--- a/twake/backend/node/src/core/platform/services/tracker/index.ts
+++ b/twake/backend/node/src/core/platform/services/tracker/index.ts
@@ -108,7 +108,7 @@ export default class Tracker extends TwakeService<TrackerAPI> implements Tracker
     callback?: (err: Error) => void,
   ): Promise<void> {
     const analiticsIdentity = {
-      userId: identity.user?.allow_tracking ? identity.user.identity_provider_id : `anonymous-${md5(identity.user.identity_provider_id)}`,
+      userId: identity.user?.allow_tracking ? identity.user.identity_provider_id : `anonymous-${md5(identity.user.identity_provider_id || "")}`,
       ...identity, //Fixme: right now we use this to send onboarding emails so user is not completely anonymous yet
     };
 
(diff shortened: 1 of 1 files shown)


twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -5,6 +5,7 @@ import TrackerAPI from "./provider";
 import { localEventBus } from "../../framework/pubsub";
 import { IdentifyObjectType, TrackedEventType, TrackerConfiguration } from "./types";
 import { ResourceEventsPayload } from "../../../../utils/types";
+import { md5 } from "src/core/crypto";
```
chamerling, line 8: src should not appear in import

twake/backend/node/src/core/platform/services/tracker/index.ts
```diff
@@ -108,7 +109,7 @@ export default class Tracker extends TwakeService<TrackerAPI> implements Tracker
   ): Promise<void> {
     const analiticsIdentity = {
       userId: identity.user?.allow_tracking ? identity.user.identity_provider_id : undefined,
-      anonymousId: "anonymous",
+      anonymousId: "anonymous-" + md5(identity.user.identity_provider_id),
```
chamerling, line 112:
``` js
`anonymous-${md5(identity.user.identity_provider_id)}`
```

twake/backend/node/src/core/platform/services/tracker/index.ts
RomaricMourgues: Right, well the tests were going to tell me soon enough!

twake/backend/node/src/core/platform/services/tracker/index.ts
BenoitTallandier: Instead of using anonymousId, could you add `anonymous-${md5(identity.user.identity_provider_id)}` to the userId property ? :)

RomaricMourgues: Ok, why is the anonymousId not usable in this case, isn't it the good way to do it usually ? (just curious)

BenoitTallandier: Segment can handle anonymousId. I guess is more used for landing analytics. 
But Amplitude can't work without userId :)

RomaricMourgues:
@BenoitTallandier Any other change to request ? I can add a new event or change events names this is your moment :p

PR #1647, opened 2021-09-21
👋 Authentication part 1

chamerling:
Replace https://github.com/linagora/Twake/pull/1563/.

New auth provider is still not connected but code is here. We now need to integrate all the parts. I will add recoil stuff once this part is integrated.


chamerling:
It is ready to merge and then I will go back to deep refactoring of the auth provider. The current PR code is mostly not used. I then need to update the login.js file

PR #1676, opened 2021-10-01
📱 Mobile deep link and fixes 01102021

RomaricMourgues:
- [x] Ensure plan features all always set in response
- [x] Add deep-link well-known files for ios and android
- [x] Verify the banner id
- [x] OnlyOffice not by default
- [x] Resolve bug happening just after registration

### DIFF
--- a/twake/frontend/public/.well-known/apple-app-site-association
+++ b/twake/frontend/public/.well-known/apple-app-site-association
@@ -1,0 +1,11 @@
+{
+    "applinks": {
+        "apps": [],
+        "details": [
+            {
+                "appID": "ABCDX.com.twake",
+                "paths": ["*"]
+            }
+        ]
+    }
+}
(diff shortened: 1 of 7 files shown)


twake/frontend/public/.well-known/assetlinks.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "relation": ["delegate_permission/common.handle_all_urls"],
+    "target": {
+      "namespace": "twake",
```
RomaricMourgues, line 5: @bobs4462 I'm not shure about this "namespace" field

twake/frontend/public/.well-known/apple-app-site-association
```diff
@@ -0,0 +1,11 @@
+{
+    "applinks": {
+        "apps": [],
+        "details": [
+            {
+                "appID": "1257653795.twake",
```
RomaricMourgues, line 6: Does it seems right to you ? @bobs4462

PR #1680, opened 2021-10-05
Allow to auth from email only in dev mode

chamerling:
@RomaricMourgues I let you the choice to merge it or not. Without it, I was not able to run the console with Twake.

### DIFF
--- a/twake/backend/node/src/services/console/web/controller.ts
+++ b/twake/backend/node/src/services/console/web/controller.ts
@@ -3,7 +3,6 @@ import { FastifyReply, FastifyRequest } from "fastify";
 import {
   AuthRequest,
   AuthResponse,
-  ConsoleExecutionContext,
   ConsoleHookBody,
   ConsoleHookBodyContent,
   ConsoleHookCompany,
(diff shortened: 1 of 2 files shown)


twake/backend/node/src/services/console/web/controller.ts
```diff
@@ -178,13 +176,21 @@ export class ConsoleController {
...
     }
-    const [storedPassword, salt] = await this.userService.users.getHashedPassword({ id: user.id });
-    if (!(await this.passwordEncoder.isPasswordValid(storedPassword, password, salt))) {
-      throw CrudExeption.forbidden("Password doesn't match");
+
+    // allow to login in development mode with any password. This can be used to test without the console provider because the password is not stored locally...
+    if (process.env.NODE_ENV !== "development") {
+      const [storedPassword, salt] = await this.userService.users.getHashedPassword({
+        id: user.id,
+      });
```
chamerling, line 187: This is the important part, other changes are just code cleaning

RomaricMourgues: Could we add some big warning logs in this function to see them if we have this enabled by mistake in one of our futures installations ?

PR #1701, opened 2021-10-14
🍺 Introduce useRealtime hook applied to workspaces

chamerling:


chamerling:
This should be OK for first version of realtime hooks, next steps about typing and presence will probably add more cool stuff in it @RomaricMourgues

PR #1748, opened 2021-11-11
🌟 Feature/magic links

romanesko:
https://user-images.githubusercontent.com/1404610/141362075-04ed1035-d439-47d9-86c4-5c6a94dc4962.mov


romanesko:
moved everything under the Skip button
Screenshot 2021-11-12 at 12 32 04

RomaricMourgues:
Great ! @BenoitTallandier I will merge it to help mobile team move forward, could you tell if this is alright ?

twake/frontend/src/app/scenes/Client/Popup/AddUser/MagicLinks.tsx
```diff
@@ -0,0 +1,120 @@
...
+    recreateToken(): Promise<MagicLinksResponse> {
+        this.loading(true);
+        return Api.post<any, { resource: MagicLinksResponse }>(this.route, {}).then(a => a.resource).finally(() => this.loading(false));
+    }
+
+    deleteToken(token: string): Promise<undefined> {
+        this.loading(true);
+        return Api.delete(this.route + '/' + token).then(a => undefined).finally(() => this.loading(false));
+    }
+}
```
RomaricMourgues, line 38: My only feedback, this class should be moved in a MagicLinksApiClient.ts file in services/workspaces
Also our convention is to use
```
`${this.route}/${…}`
```
Instead of string concatenation with « + »

RomaricMourgues (changes requested):

twake/frontend/src/app/scenes/Client/Popup/AddUser/MagicLinks.tsx
romanesko: If you allow me to disagree — I would do this! )
The one of the coolest things of react — it's a component isolation. It's ok to have services outside of component is the case of multiple usage. But data related to one component and used only on one component is even better to keep inside of this component (as far as framework supports and approve this).

But last word is yours, anyway.

Concatenation fix coming soon.

RomaricMourgues:
I agree, it means that we have API logic both in components and in services depending on feature.
How can we avoid this ? There is some cleanness to have all the hooks (useSomeyhing) in the same folder for reference for instance.
Just take this in consideration and do what you prefer, it is no big work to change in the future if we want to refactor this anyway :)

twake/frontend/src/app/scenes/Client/Popup/AddUser/MagicLinks.tsx
RomaricMourgues: I agree, it means that we have API logic both in components and in services depending on feature.
How can we avoid this ? There is some cleanness to have all the hooks (useSomeyhing) in the same folder for reference for instance.
Just take this in consideration and do what you prefer, it is no big work to change in the future if we want to refactor this anyway :)

Issue #1750, opened 2021-11-12
Do not load all the workspaces when opening Twake


RomaricMourgues:
**Describe the bug**
- [x] Currently all the workspaces are loaded one by one across all the companies of the user. It slow down the startup of the app.
- [x] Also when loading list of workspaces from a company, do not return workspaces I do not belong to
- [x] We need to auto-load last used workspace

**To Reproduce**
Open Twake having at least 2 companies (every workspaces of every company will be loaded)

**Expected behavior**
Load only what is necessary to open Twake as quickly as possible.

**Additional context**
Options considered:
- Only load the workspaces of the currently visible company
- Add a single url to load all the workspaces at once from the backend (it seems more clean to load all the workspace but still per company, not across all companies because it seems anti-pattern)

PR #1764, opened 2021-11-23
⚡️ #1750 Do not load all the workspaces when opening Twake

stephanevieira75:
ref to #1750


RomaricMourgues:
@stephanevieira75 I did a lot of changes:
- Now we have a Company atom with all the companies in it
- We have a User atom
- We use the differents hooks in the correct order: useCurrentUser, useCurrentCompany, useCompanyWorkspaces (to make sure the previous one is always loaded
- I added the No Workspaces and the No Companies pages
- I fixed the auto-load of companies (but not the auto-load of workspaces from best candidate)

Issue #1790, opened 2021-12-03
Implement realtime for the current user in frontend


stephanevieira75:
The goal is to make sure user information (first name, language, timezone, avatar etc) is updated in realtime without needing to reload the page.

For this we will create a `/me/{user-id}` realtime room and each time the user entity is updated (in the user service) we should send the user entity in this room.

When the frontend receive an event from this room, it should refresh the current user completely.

This should work for any of this cases:
- first name / last name / avatar / status update
- companies list of the user update
- language / timezone update
- any preferences of the user is updated

This websocket room will be protected only by the user's jwt that contains the user id (user can only access it's own room of course).

RomaricMourgues:
Not working on canary server (Prod console <-> Canary Twake issues ?)

PR #1815, opened 2021-12-14
🌟 #1790 Implement realtime for user, company and applications

stephanevieira75:


RomaricMourgues:
@stephanevieira75 your error is 
```
{"level":50,"time":1639578909239,"pid":405,"hostname":"d50c885c969e","reqId":"req-b","stack":"TypeError: Cannot read properties of undefined (reading 'id')\n    at UsersCrudController.getCompany (/usr/src/app/src/services/user/web/controller.ts:211:31)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)","type":"Error","msg":"Cannot read properties of undefined (reading 'id')"}
```

twake/backend/node/src/services/user/services/companies/service.ts
```diff
@@ -60,6 +61,15 @@ export class CompanyService implements CompaniesServiceAPI {
...
 
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  @RealtimeSaved<Company>((company, _context) => {
+    return [
+      {
+        room: `/company/${company.id}`,
+        resource: company,
+      },
+    ];
+  })
```
RomaricMourgues, line 72: This will not work, the updateCompany function must contain a "context" as last argument and should return a SaveResult object.

RomaricMourgues: Also we need to add a RealtimeSaved inside removeUserFromCompany and setUserRole (to make companies appear or disappear)

RomaricMourgues (changes requested):

twake/backend/node/src/services/user/services/companies/service.ts
stephanevieira75: Updated, could you check if it's good please? :)

Issue #1813, opened 2021-12-14
Messaging for the section to be updated


shivam9593:
![image.png](https://images.zenhubusercontent.com/619e1750e7e4f47b0d4a07ff/8c5e3d6d-a4c0-4b1c-af73-0961c79cf0b8)

Change the messaging from "Start writing the name of your first channel section" to "You can group your channels here"

Issue #1812, opened 2021-12-14
Emojis in the channel creation are too small


shivam9593:
Emojis in the channel creation are too small for someone to see. 

Is it possible to maybe increase the size of them a bit for better user experience. 

![image.png](https://images.zenhubusercontent.com/619e1750e7e4f47b0d4a07ff/d5ea32f7-f164-40e5-b786-67f790528253)

RomaricMourgues:
Done
![Screenshot 2021-12-14 at 13 44 14](https://user-images.githubusercontent.com/14353924/146000657-791ae5a9-071b-46b1-a8b4-07bf953a4aa8.png)

PR #1816, opened 2021-12-14
🌟 Messages in node part 4

RomaricMourgues:
- [x] Real-time for messages
- [x] Input slow to write
- [x] Sending only a file
- [x] Twacode to put back
- [x] Sending a message with a in-progress upload
- [x] Still not ok the retrieval of list of companies from console on QA
- [x] Applications search not working (migration to do ? => auto load services even if not added in tree)
- [x] #1813
- [x] #1812

Part 5
- Change tab
- Ephemeral messages
- Files / pinned / bookmark views API
- Upload attachments from drive
- Mentions


twake/backend/node/src/services/files/web/controllers/files.ts
```diff
@@ -48,6 +48,7 @@ export class FileController {
     const data = await this.service.download(params.id, context);
     const filename = data.name.replace(/[^a-zA-Z0-9 -_.]/g, "");
 
+    response.header("Access-Control-Allow-Headers", "*");
```
tanandy, line 51: can it be configurable ? for security reason

RomaricMourgues: I removed it, not needed anymore

Issue #1811, opened 2021-12-14
Re-implement the "is writing" event


RomaricMourgues:
**Functional**

Add a small text with the names of the currently writing users under the list of messages (in channel and in thread view):
- Version with two users: "Romaric, Benoit are writing..."
- Version with one user: "Romaric is writing..."

We can have a little animation but not too much, we cannot have a bubble like in messenger because the design is too different (we can take inspiration from discord, slack and teams).

The channel should have a little writing icon when someone is writing in it on the channel bar.

**Technical detail**

This realtime event will goes through the workspace websocket room (this will allow us to see the channel being edited in realtime and prepare for the call system update later).

**Tasks**

- [ ] Create an atomFamily named "ChannelWritingActivityState" with family key "channelId" and value `{threadId, userId, name, isWriting: boolean}[]`
- [ ] Create a selector to listen to only a threadId
- [ ] Create a useChannelWritingActivity(channelId, threadId: string | null) => {users: {userId, name}[], iAmWriting: Function}

PR #1823, opened 2021-12-15
 🛠 #1811:  Re-implement the "is writing" event

tissarni:
Re-implement the feature that allow users to know if someone is writing in Twake

## Description

- [x]  Create an atomFamily named "ChannelWritingActivityState" with family key "channelId" and value {threadId, userId, name, isWriting: boolean}[]
- [x]  Create a selector to listen to only a threadId
- [x]  Create a useChannelWritingActivity(channelId, threadId: string | null) => {users: {userId, name}[], iAmWriting: Function}

- [x] Use ThreadWritingActivity selector to add "is writing info" in thread view 
- [x] Create a component to be use in the channel bar to show that someone is writing in a channel 

- [x] Don't use input onChange method to call "IAmWriting()" and find a way to call it only when user type in the editor. (onChange method is called to much time even if the editor don't change). 

- [x] Add doc about the websocket room and event format
- [x] Filter "is writing" info for the user writing 
- [x] Add translation, now only English is availaible 
- [x] To change design, look here: 
          - frontend/src/app/scenes/Client/ChannelsBar/Parts/Channel/Channel.scss
          - frontend/src/app/components/WritingLoader

## Related Issue

https://github.com/linagora/Twake/issues/1811

### DIFF
--- a/twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
+++ b/twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
@@ -1,15 +1,12 @@
 import { atomFamily } from 'recoil';
 
-export type AtomChannelIdKey = {
+export type ChannelWritingActivityType = {
   threadId: string;
   userId: string;
   name: string;
-  isWritting: boolean;
 };
 
-export const ChannelWrittingActivityState = atomFamily<AtomChannelIdKey[], AtomChannelIdKey>({
-  key: 'ChannelWrittingActivityState',
-  default: AtomChannelIdKey => {
-    return {};
-  },
+export const ChannelWritingActivityState = atomFamily<ChannelWritingActivityType[], string>({
+  key: 'ChannelWritingActivityState',
+  default: [],
 });
--- a/twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
+++ b/twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
@@ -1,0 +1,15 @@
+import { atomFamily } from 'recoil';
+
+export type AtomChannelIdKey = {
+  threadId: string;
+  userId: string;
+  name: string;
+  isWritting: boolean;
+};
+
+export const ChannelWrittingActivityState = atomFamily<AtomChannelIdKey[], AtomChannelIdKey>({
+  key: 'ChannelWrittingActivityState',
+  default: AtomChannelIdKey => {
+    return {};
+  },
+});
--- a/twake/frontend/src/app/state/recoil/hooks/useChannelWrittingActivity.ts
+++ b/twake/frontend/src/app/state/recoil/hooks/useChannelWrittingActivity.ts
@@ -1,6 +1,0 @@
-export default function useChannelActivityWriting(
-  chanelId: string,
-  threadId: string | null,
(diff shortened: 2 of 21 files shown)


twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
```diff
@@ -0,0 +1,15 @@
+import { atomFamily } from 'recoil';
+
+export type AtomChannelIdKey = {
+  threadId: string;
+  userId: string;
+  name: string;
+  isWritting: boolean;
+};
+
+export const ChannelWrittingActivityState = atomFamily<AtomChannelIdKey[], AtomChannelIdKey>({
```
RomaricMourgues, line 10: No, here the family key is a simple string that will contain the channel_id.

twake/frontend/src/app/state/recoil/hooks/useChannelWrittingActivity.ts
```diff
@@ -0,0 +1,6 @@
+export default function useChannelActivityWriting(
+  chanelId: string,
+  threadId: string | null,
+): { users: { userId: string; name: string }[]; iAmWritting: any } {
```
RomaricMourgues, line 4: You can create a type for the return.

twake/frontend/src/app/state/recoil/selectors/ThreadWrittingActivity.ts
```diff
@@ -0,0 +1,7 @@
+import { selector } from 'recoil';
+import { ChannelWrittingActivityState } from '../atoms/ChannelWritingActivity';
+
+export const ThreadWrittingActivitySelector = selector<string>({
+  key: 'ThreadWrittingActivitySelector',
+  get: ({ get }) => get(ChannelWrittingActivityState)?.threadId || '',
+});
```
RomaricMourgues, line 7: Not exactly, the goal here is to listen only to the thread, it looks like this:

```
export const ThreadWrittingActivitySelector = selectorFamily<{channelId: string, threadId: string}>({
  key: 'ThreadWrittingActivitySelector',
  get: ({threadId, channelId}) => ({ get }) => (get(ChannelWrittingActivityState(channelId)) || []).filter(status => status.threadId === threadId),
});
```

(not tested but you'll get the idea)

RomaricMourgues (changes requested):

twake/frontend/src/app/state/recoil/atoms/ChannelWritingActivity.ts
RomaricMourgues: ChannelWrittingActivityStateType

twake/frontend/src/app/state/recoil/selectors/ThreadWrittingActivity.ts
RomaricMourgues: Even better: 
```
export const ThreadWrittingActivitySelector = selectorFamily<{channelId: string, threadId: string}>({ key: 'ThreadWrittingActivitySelector', get: ({threadId, channelId}) => ({ get }) => (get(ChannelWrittingActivityState(channelId)) || []).filter(status => status.isWriting && status.threadId === threadId), });
```

RomaricMourgues:
@tissarni could you close the resolved discussions and check the finished tasks (and add new tasks if you have new steps)

PR #1826, opened 2021-12-15
🌟 Applications node part 1

romanesko:
it's a draft


RomaricMourgues:
@romanesko ping me when it's ready to merge for step 2, we'll create a new PR for step 3 and later

romanesko:
It's ready
But I need to notice
1. now responses of /api/* endpoints doesn't return `resource` (regarding the docs), it's could be a mistake and we need to wrap it.
2. the privateKey returns only once right after the application creating (not after get or edit)

RomaricMourgues:
For the 1. ok we'll need to fix it later still
For the 2. are we talking about the app secret ? Why is it visible only once ?

romanesko:
1. I can fix it now, that's quick
2. we have this data available to any admin of the company, it does not look very right, but I can open the data for all of them

RomaricMourgues:
Yes for now it will be visible to all admin in the company I confirm (but good point I will keep it in mind)

RomaricMourgues:
@romanesko ping me by sms or Twake when my last comment is implemented so I can merge :)

Issue #1844, opened 2021-12-22
Fix Calendar event creation


stephanevieira75:

RomaricMourgues:
Tested multiple time on canary and QA

PR #1845, opened 2021-12-22
#1844 Fix calendar event creation

stephanevieira75:
https://user-images.githubusercontent.com/36481167/147086301-dd4a716f-777b-4093-9a13-3f12066c7371.mp4


RomaricMourgues:
Ok thanks, so it decrease the performances a bit but that's ok for old code base 👍

Issue #1861, opened 2022-01-03
Fix onboarding tour to show only once


BenoitTallandier:
Currently, onboarding tour saves his status in the local storage. When a user use an other browser, the tour in shown again. 

Usetiful just shipped a new feature for that. We just need to add userId in the usetiful tag.
For that, we need to :
1. add te usetiful tag in the code (for now, it is imported by google tag manager).
2. add the userId tag.

![image.png](https://images.zenhubusercontent.com/5f06c5a0e15425885ef10c4a/74c47505-e650-4fed-9fac-8cdef85d2916)

usetiful code : 
```
<!-- Usetiful script start -->
<script>   
  (function (w, d, s) {
    var a = d.getElementsByTagName('head')[0];
    var r = d.createElement('script');
    r.async = 1;
    r.src = s;
    r.setAttribute('id', 'usetifulScript');
    r.dataset.token = "2bfe9f6de8087d492cd66e4760b2fade";
            a.appendChild(r);
  })(window, document, "https://www.usetiful.com/dist/usetiful.js");
  
</script>
<!-- Usetiful script end -->

```

RomaricMourgues:
@BenoitTallandier not sure this is working on qa and canary

BenoitTallandier:
I confirm, It's not working on QA.

hantt12:
Tested in QA and Production today and there are no Onboarding Tour?

PR #1902, opened 2022-01-26
#1861 Fix onboarding tour to show only once

stephanevieira75:


twake/frontend/src/app/environment/types.ts
```diff
@@ -7,6 +7,7 @@ export type EnvironmentType = {
   websocket_url: string;
 
   sentry_dsn?: string | boolean;
+  usetiful_token: string | undefined;
```
RomaricMourgues, line 10: This should be optional

RomaricMourgues (changes requested):

twake/frontend/src/app/environment/types.ts
stephanevieira75: Done

RomaricMourgues:
@BenoitTallandier could you test when it will be on qa (tag staging:qa)

Issue #2007, opened 2022-03-10
Improve messages search backend


RomaricMourgues:
- [x] Filter out messages from channels we are not member of (see at the end)
- [ ] Add options for search: `has_files true / false` and `sender: uuid`. has_files must be computed on messages.search.ts when defining the source like this: `has_files: (entity.files || []).length > 0`
- [x] Update tests
- [ ] Add/Complete documentation on Notion page (at then very end) https://www.notion.so/linagora/Rest-APIs-232eb8e51c3743ccbb7aa3534a5a7db9 (`/companies/{company_id}/search`)

About filtering out messages, line 119 of twake/backend/node/src/services/messages/web/controllers/views.ts :
```
    //TODO for each message check we have access to it (check we are member in the channel of this message + avoid making everything slow while doing this check so maybe some quick cache)
    //TODO as some messages will be filtered out after previous TODO, we should loop on the elastic search calls to reach the expected limit (For instance first 100 messages returned by ES could all be from unaccessible channel so we need to get the next 100 messages and so on)
```

PR #2012, opened 2022-03-15
Search improve

romanesko:
https://github.com/linagora/Twake/issues/2007

### DIFF
--- a/twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
+++ b/twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
@@ -257,7 +257,6 @@ export default class ElasticSearch extends SearchAdapter implements SearchAdapte
       esResponse = await this.client.scroll(
         {
           scroll_id: options.pagination.page_token,
-          // ...esParamsWithScroll
         },
         esOptions,
       );
--- a/twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
+++ b/twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
@@ -28,7 +28,7 @@ type Operation = {
 
 export default class ElasticSearch extends SearchAdapter implements SearchAdapterInterface {
   private client: Client;
-  private bulkReaders: number = 0;
+  private bulkReaders = 0;
   private buffer: Operation[] = [];
   private name = "ElasticSearch";
 
@@ -172,7 +172,7 @@ export default class ElasticSearch extends SearchAdapter implements SearchAdapte
       return;
     }
 
-    logger.info(`Start new Elasticsearch bulk reader.`);
+    logger.info("Start new Elasticsearch bulk reader.");
     this.bulkReaders += 1;
 
     let buffer;
@@ -229,7 +229,7 @@ export default class ElasticSearch extends SearchAdapter implements SearchAdapte
       logger.error(err);
     }
 
-    logger.info(`Elasticsearch bulk flushed.`);
+    logger.info("Elasticsearch bulk flushed.");
     this.bulkReaders += -1;
 
     this.startBulkReader();
@@ -252,9 +252,13 @@ export default class ElasticSearch extends SearchAdapter implements SearchAdapte
     };
 
     let esResponse: ApiResponse;
+
     if (options.pagination.page_token) {
       esResponse = await this.client.scroll(
-        { scroll_id: options.pagination.page_token, ...esParamsWithScroll },
+        {
+          scroll_id: options.pagination.page_token,
+          // ...esParamsWithScroll
+        },
         esOptions,
       );
     } else {
--- a/twake/backend/node/src/services/channels/services/member/service.ts
+++ b/twake/backend/node/src/services/channels/services/member/service.ts
@@ -47,6 +47,7 @@ import { CounterProvider } from "../../../../core/platform/services/counter/prov
 import { PlatformServicesAPI } from "../../../../core/platform/services/platform-services";
 import { countRepositoryItems } from "../../../../utils/counters";
 import { getService as getCompanyService } from "../../../user/services/companies";
+import NodeCache from "node-cache";
 
 const USER_CHANNEL_KEYS = [
   "id",
@@ -78,6 +79,7 @@ export class Service implements MemberService {
   channelMembersRepository: Repository<MemberOfChannel>;
   private channelCounter: CounterProvider<ChannelCounterEntity>;
   private companies: CompaniesServiceAPI;
+  private cache: NodeCache;
 
   constructor(
     private platformServices: PlatformServicesAPI,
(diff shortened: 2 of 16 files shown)


twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
```diff
@@ -252,9 +252,13 @@ export default class ElasticSearch extends SearchAdapter implements SearchAdapte
     };
 
     let esResponse: ApiResponse;
+
     if (options.pagination.page_token) {
       esResponse = await this.client.scroll(
-        { scroll_id: options.pagination.page_token, ...esParamsWithScroll },
+        {
+          scroll_id: options.pagination.page_token,
+          // ...esParamsWithScroll
```
RomaricMourgues, line 260: Could you explain to me why is it removed?
And then depending on the response I suppose there is two option here:
- Put it back
- Remove it completely
😛

twake/backend/node/src/services/messages/services/views/service.ts
```diff
@@ -222,6 +222,7 @@ export class ViewsService implements MessageViewsServiceAPI {
         ...(options.companyId ? { $in: [["company_id", [options.companyId]]] } : {}),
         ...(options.workspaceId ? { $in: [["workspace_id", [options.workspaceId]]] } : {}),
         ...(options.channelId ? { $in: [["channel_id", [options.channelId]]] } : {}),
+        // TODO: implement boolean for hasFiles
```
RomaricMourgues, line 225: You can already do it:

```
async search(
    pagination: Pagination,
    options?: SearchMessageOptions,
    context?: ExecutionContext,
  ): Promise<ListResult<Message>> {
    return await this.searchRepository.search(
      {
      //Here:
        has_file: myBoolean
      },
      {
        pagination,
        ...(options.companyId ? { $in: [["company_id", [options.companyId]]] } : {}),
        ...(options.workspaceId ? { $in: [["workspace_id", [options.workspaceId]]] } : {}),
        ...(options.channelId ? { $in: [["channel_id", [options.channelId]]] } : {}),
```

RomaricMourgues: Same place for the `sender`

twake/backend/node/src/services/messages/web/controllers/views.ts
```diff
@@ -1,26 +1,65 @@
...
+    return this.data[key];
+  }
+
+  set(key: string, value: { [key: string]: boolean }) {
+    this.expiration[key] = SimpleChannelsCache.now() + this.minutesTTL * 1000;
+    this.data[key] = value;
+  }
+}
+
+const simpleChannelsCache = new SimpleChannelsCache(5);
```
RomaricMourgues, line 54: Actually we already have a cache available `import NodeCache from "node-cache";` you can see it in action here: twake/backend/node/src/services/user/services/users/service.ts

RomaricMourgues (changes requested):
Nice to have you back on the code!

I have some feedback, that's normal after this delay. There is mainly small improvements asked on my comments.

I also need to discuss with you about the current strategy of getting all the channels first and then looking at this big list on the search results. Take a look at how I did it for listing messages and adding the include_users in it: `includeUsersInDirectChannel` in `src/services/channels/services/channel/service.ts`

Thanks for the awesome work on tests and global structure of the code!

twake/backend/node/src/core/platform/services/search/adapters/elasticsearch/index.ts
romanesko: It doesn't work with query in `_search/scroll`. It expects only `scroll_id` that refers to the previous query, and it falls with error if query exists in this request. I tested on a local elastic search and found this pattern. The documentation mentions it in passing, but it doesn't work with `query` anyway. So I suppose we can remove it completely.

twake/backend/node/src/services/messages/web/controllers/views.ts
romanesko: 🤦‍♂️ … 
mine is pretty elegant too )
ok, I'll change it to node-cache

twake/backend/node/src/services/messages/services/views/service.ts
romanesko: Couldn't handle with has_files actually, need your help.

RomaricMourgues:
Yes it was pretty elegant unfortunately 😬

RomaricMourgues:
Interesting ! Scroll_id probably contain the information of pagination and stuff already I suppose. Thanks for the information you can remove it then ;)

RomaricMourgues:
If you can share the generated elastic search query which should appear in the logs I can probably help
I mean if the problem is in the elastic search query of course

twake/backend/node/src/services/messages/entities/messages.search.ts
```diff
@@ -24,9 +26,11 @@ export default {
   esMapping: {
     properties: {
       text: { type: "text" },
+      user_id: { type: "text" },
```
RomaricMourgues, line 29: If you want to search using this format:
`...(options.sender ? { $in: [["user_id", [options.sender]]] } : {}),`
You need to use this type:
`user_id: { type: "keyword" }`

twake/backend/node/src/services/channels/services/member/service.ts
```diff
@@ -583,17 +618,32 @@ export class Service implements MemberService {
...
-      workspace_id: channel.workspace_id,
-      user_id: user.id,
-    });
+    const get = () =>
+      this.get({
+        channel_id: channel.id,
+        company_id: channel.company_id,
+        workspace_id: channel.workspace_id,
+        user_id: user.id,
+      });
```
RomaricMourgues, line 636: I don't think we still need a function in a function, you can move this code block to the end return (the variable `const get` is only used once

twake/backend/node/src/services/messages/web/controllers/views.ts
```diff
@@ -113,27 +114,73 @@ export class ViewsController {
...
-        context,
-      )
-      .then(a => a.getEntities());
+        50,
+      );
+      if (!isChannelMember) continue;
+
+      // if (request.query.sender && msg.user_id != request.query.sender) continue;
+
+      // if (request.query.has_files && !(msg.files || []).length) continue;
```
RomaricMourgues, line 171: This comments should be removed when tests pass I suppose :)

twake/backend/node/src/services/messages/entities/messages.search.ts
romanesko: hm… but it works this way…
ok, I'll change

RomaricMourgues: I didn't know it worked haha :p 
But another point for me is that everything you put in "text" will be searchable in the "query" like if the message text is "I love the option b514" and sender is "0ac1d15a-a5dd-11ec-b514-476f788a5663" then searching for the text "b514" will return all the messages from this sender.
So change it to `keyword` anyway ;)

RomaricMourgues:
For code it is ok for me, but tests don't like it yet :)

romanesko:
Is it possible that tests use non-elastic-search configuration? Can't figure out

PR #2023, opened 2022-03-22
🛠 global resolver

romanesko:
it's a draft of global reserver


twake/backend/node/src/services/global-resolver.ts
```diff
@@ -0,0 +1,153 @@
...
+  workspaces: WorkspaceService;
+  companies: CompaniesServiceAPI;
+  users: UsersService;
+  console: ConsoleServiceAPI;
+  statistics: StatisticsAPI;
+  externalUser: UserExternalLinksService;
+  notificationBadge: UserNotificationBadgeServiceAPI;
+  notificationPreferences: UserNotificationPreferencesAPI;
+  messages: MessageThreadMessagesServiceAPI;
+  threads: MessageThreadsServiceAPI;
```
RomaricMourgues, line 76: Why not just import MessageServiceAPI ? It contains:
```
  userBookmarks: MessageUserBookmarksServiceAPI;
  threads: MessageThreadsServiceAPI;
  messages: MessageThreadMessagesServiceAPI;
  views: MessageViewsServiceAPI;
  engine: MessagesEngine;
```

twake/backend/node/src/services/global-resolver.ts
```diff
@@ -0,0 +1,153 @@
...
+  userBookmarks: MessageUserBookmarksServiceAPI;
+  applications: MarketplaceApplicationServiceAPI;
+  companyApplications: CompanyApplicationServiceAPI;
```
RomaricMourgues, line 79: Same here, ApplicationServiceAPI contains already this MarketplaceApplicationServiceAPI and CompanyApplicationServiceAPI

twake/backend/node/src/services/global-resolver.ts
```diff
@@ -0,0 +1,153 @@
...
+  views: MessageViewsServiceAPI;
+  messageEngine: MessagesEngine;
+  files: FileServiceAPI;
+  channels: ChannelService;
+  members: MemberService;
+  channelPendingEmail: ChannelPendingEmailService;
+  tab: TabService;
+};
```
RomaricMourgues, line 87: In general, I think we would get better code if we have one item here for each folder we have in /services/ or /core/services/

twake/backend/node/src/services/global-resolver.ts
```diff
@@ -0,0 +1,153 @@
...
+type PlatformServices = {
+  realtime: RealtimeServiceAPI;
+  auth: AuthServiceAPI;
+  search: SearchServiceAPI;
+  storage: StorageAPI;
+  pubsub: PubsubServiceAPI;
+  counter: CounterAPI;
+};
+
+type CoreServices = {
```
RomaricMourgues, line 66: The name would be more like "TwakeServices" as core is the name of the folder in which "PlatformServices" are found.

RomaricMourgues (changes requested):
So three topics:
- Rename the _services in config
- Rename CoreServices in global-resolver
- Reduce number of services in global-resolver

Also I am thinking that maybe we could have an other global-resolver for the platform services so in `/core/platform/services/global-resolver.ts` instead of having PlatformServices in the current global-resolver.

twake/backend/node/config/default.json
```diff
@@ -139,6 +143,20 @@
...
+    "pubsub",
+    "push",
+    "realtime",
+    "search",
+    "storage",
+    "tracker",
+    "webserver",
+    "websocket"
+  ],
+  "_services": [
```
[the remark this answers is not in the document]
romanesko: Yes, it should be removed

twake/backend/node/src/services/global-resolver.ts
romanesko: There is no more MessageServiceAPI  😬 
One of goals was to remove this grouping services, because of the over-use of the same services in different groups.

I told you today that I'm going to group some of them on level of global resolver  for convinience

twake/backend/node/src/services/global-resolver.ts
romanesko: Not sure if I understand you right, but we have here one item for each folder we have in services…. no?

RomaricMourgues: We don't have a /services/channelPendingEmail nor a /services/threads folders, it is related to the "I told you today that I'm going to group some of them on level of global resolver for convinience"

twake/backend/node/config/default.json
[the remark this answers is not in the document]
romanesko: unfortunately we need all services here for now, after I change the way web services are detected it will be necessary to specify only the platform services here

romanesko:
I've pushed fixes, take a look

RomaricMourgues:
@romanesko why is this not merged yet ? The more we work on other stuff the more difficult it will be to merge right ? :p

romanesko:
> @romanesko why is this not merged yet ?
I can't merge, only you can

Issue #1912, opened 2022-01-27
Implement puppeteer and start testing the frontend


stephanevieira75:
https://www.npmjs.com/package/puppeteer

PR #2101, opened 2022-04-20
🌟 #1912 Implement puppeteer and start testing frontend

stephanevieira75:


RomaricMourgues:
That seems nice ! Could you create a `docker-compose.tests.integration.yml` that will run your tests from `.github/workflows` ?

RomaricMourgues:
@stephanevieira75 2 tests out of 3 are passing, the create channel one is failing.
Also "this is the config file" log should be removed.

Issue #2071, opened 2022-04-13
Build blocks<->twacode converter


RomaricMourgues:
Today Twake web is using "twacode" which are deprecated.
We initiated the next version called "Blocks" that follows this documentation from Slack: https://api.slack.com/block-kit
As mobile app will soon start the work on the plugins, we need to make sure they start working with the new Block api and not the old one.

**To Do**: build a both ways converter for Twacode and Blocks.

**How to do it**:

1. All the types of the new format are in here: `twake/backend/node/src/services/messages/blocks-types.ts` first step will be to duplicate this in a small node project
2. We will build a very small and simple test node project in typescript that will define two methods: `twacodeToBlocks(twacode: any)` and `blocksToTwacode(blocks: Block[])`
3. Take time to understand difference in Slack API between Blocks (top level blocks) and BlockElements (nested blocks)
4. You'll be able to test conversion both ways with this slack tool: https://app.slack.com/block-kit-builder

How to build the code (WIP)

PR #2109, opened 2022-04-22
🌟 2071: Create BlockCompiler + update blocks-type and use block compiler

tissarni:
Creating a block compiler to be able to use Slacks block and compile the block into Twacode to use it in Twake 

## Related Issue

https://github.com/linagora/Twake/issues/2071 : Build blocks<->twacode converter

### DIFF
--- a/twake/frontend/src/app/components/twacode/blocksCompiler.ts
+++ b/twake/frontend/src/app/components/twacode/blocksCompiler.ts
@@ -118,7 +118,7 @@ export const blocksToTwacode = (event: Block[]) => {
       // TODO
     }
   });
-  return [{ type: 'twacode', elements: blocks }];
+  return blocks;
 };
 
 const createTwacodeElements = (object: any, blocks: any[], type?: any) => {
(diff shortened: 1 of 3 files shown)


twake/frontend/src/app/components/twacode/blocksCompiler.ts
```diff
@@ -0,0 +1,264 @@
...
+        );
+        break;
+      case 'twacode':
+        blocks.push(object.elements);
+        break;
+      default:
+      // TODO
+    }
+  });
+  return [{ type: 'twacode', elements: blocks }];
```
RomaricMourgues, line 121: We can return directly "blocks" here and... (see next comment)

twake/frontend/src/app/views/applications/messages/message/parts/Blocks.tsx
```diff
@@ -76,7 +81,7 @@ export default React.memo((props: Props) => {
   return (
     <Suspense fallback={<></>}>
       <Twacode
-        content={props.blocks?.[0]?.elements || []}
+        content={blocks[0].elements}
```
RomaricMourgues, line 84: (see previous comment) ... and directly put content={blocks} here.

RomaricMourgues (changes requested):

PR #2180, opened 2022-05-20
🌟 2164 recent channels

romanesko:


twake/backend/node/src/services/channels/web/controllers/channel.ts
```diff
@@ -486,6 +487,66 @@ export class ChannelCrudController
...
+    });
+
+    const activities = await Promise.all(
+      res.map(channel => gr.services.channels.channels.getChannelActivity(channel)),
+    );
+
+    const response: ChannelObject[] = [];
+
+    for (let i = 0; i < res.length; i++) {
+      const channel = res[i];
```
RomaricMourgues, line 538: Just for your info this is good stuff:
```
for (const channel of res){
 ...
}
```

romanesko: I know, but I need an index there, because I merge two arrays

Issue #1519, opened 2021-08-03
[Workspace Setting] There is no a successful notification message when changing the workspace name or uploading the logo successfully


vttuann:
**Describe the bug**

Currently, If the user changes the workspace name or uploads the logo successfully, but there is not a message to notify the user that they have successfully updated.

**To Reproduce**
1. Go to 'Workspace Setting'
2. Changing the name of workspace or upload a logo successfully 
3. Verify issue

**Expected behavior**

After changing the workspace name or uploading the logo successfully, nothing happens when I click on the Update button. This means that the user does not know if they have successfully updated or not. For me, we should give a success message in this case.

**Screenshots**

![Peek 2021-08-03 10-15](https://user-images.githubusercontent.com/69950706/127952781-32734c1e-ed1e-4926-8acd-490aad24076a.gif)

**Desktop (please complete the following information):**

- OS: Ubuntu 18.06 LTS
- Browser [Chrome, Firefox]
- Version [90, 88]

**Enviroment**
Tested on QA Environment

RomaricMourgues:
Technical notes:
- You should use the existing toaster library to display notification `import { ToasterService } from 'app/features/global/services/toaster-service';`
- Make sure this is not already fixed

PR #2290, opened 2022-06-14
🛠  Fix notification message when updating workspace infos

oubchid:
## Description
- [x] Fix delete logo button 
- [x] Add notification message when updating workspace name or uploading the logo.

## Related Issue

#1519 

## Motivation and Context
Give feedback to the user when update successful done

## How Has This Been Tested?
localy

## Screenshots (if appropriate):

* Fix delete logo button 

https://user-images.githubusercontent.com/7097550/173547408-c1a10b5b-61d6-4208-94d3-ec3ef1dd0826.mp4

* Fix delete logo button

https://user-images.githubusercontent.com/7097550/173606386-d22585df-6e44-4395-a1c3-6d214df1a25f.mp4

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added the Signed-off-by statement at the end of my commit message.

### DIFF
--- a/twake/backend/node/src/services/workspaces/services/workspace.ts
+++ b/twake/backend/node/src/services/workspaces/services/workspace.ts
@@ -198,7 +198,7 @@ export class WorkspaceServiceImpl implements WorkspaceService {
     }/workspaces/${workspace.id}/thumbnail?t=${new Date().getTime()}`;
     let logoPublicUrl = undefined;
     if (workspace.logo) {
-      if (!item.logo || options.logo_b64) {
+      if (!options.logo_b64) {
         await gr.platformServices.storage.remove(logoInternalPath);
         workspace.logo = null;
       }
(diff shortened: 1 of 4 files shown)


twake/backend/node/src/services/workspaces/services/workspace.ts
```diff
@@ -198,7 +198,7 @@ export class WorkspaceServiceImpl implements WorkspaceService {
     }/workspaces/${workspace.id}/thumbnail?t=${new Date().getTime()}`;
     let logoPublicUrl = undefined;
     if (workspace.logo) {
-      if (!item.logo || options.logo_b64) {
```
RomaricMourgues, line 201: Mmm could you check that it doesnt get triggered when we change only the name of the workspace for instance ?
All our APIs are made in a way we can send only partial update. So if we send partial update with only the name and not the logo, then the logo will be deleted.
A good alternative could be to add `(!item.logo && item.logo !== undefined)` Like this sending an empty string will remove the logo.

RomaricMourgues (changes requested):

twake/backend/node/src/services/workspaces/services/workspace.ts
oubchid: item is usless here, item.logo is always filled here we cause we always get the workspace infos including `item.logo`.
When you press the delete btn to delete the logo item is not affected only options.logo_b64 is updated

RomaricMourgues (approved):

PR #2348, opened 2022-06-23
👨‍💻 molecule/media

romanesko:


twake/frontend/src/app/molecules/media/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  size: MediaSize;
+  url: string;
+  duration?: string;
+}
+
+export enum MediaSize {
+  LG = 32,
+  MD = 16,
+  SM = 12,
+}
```
RomaricMourgues, line 15: On other components I put "sm" "md" "lg", could we use the same ? (I don't mind having a enum that's ok)

romanesko: So are you want me to move to strings, or just me to create the universal Enum for all the atoms/molecules?

romanesko:
Screenshot 2022-06-23 at 17 51 08

twake/frontend/src/app/molecules/media/index.tsx
RomaricMourgues: Move to string

PR #2375, opened 2022-06-30
🧹 removing CRUD Interface and provider's interfaces

romanesko:
This PR is about:
- Make context optional
- Removing CRUD interface


romanesko:
@RomaricMourgues  yes, right now I'm just getting rid of the interface, but we can save the context wherever we need (and make it universal)

Issue #2451, opened 2022-08-10
Message delivery status


RomaricMourgues:

tprudentova:
Hi @rezk2ll! Here are some bugs I noticed
1.	In a direct chat with oneself all messages sent recently are marked as unread. The older messages are marked as read though
![Screenshot 2022-09-02 at 09.26.36.png](https://images.zenhubusercontent.com/61e03d0c21ae7b3b7c6d837c/8c7a9286-8ca2-4f8f-a26d-995be702c4c6)
2.	When you press Information, it always states that no one has read the message yet, despite the delivery status 
![Screenshot 2022-09-02 at 09.12.33.png](https://images.zenhubusercontent.com/61e03d0c21ae7b3b7c6d837c/9c58ffba-4ded-4f68-b4ce-650d4eab5a29)
3.	I believe the message is supposed to be marked as read (two blue checks) when the recipient opens the chat. However, at the moment it is marked as read even if the recipient is simply online 
![Screenshot 2022-09-02 at 09.36.45.png](https://images.zenhubusercontent.com/61e03d0c21ae7b3b7c6d837c/296df925-1441-4021-9214-e0c16b0c3cae)
![Screenshot 2022-09-02 at 09.36.59.png](https://images.zenhubusercontent.com/61e03d0c21ae7b3b7c6d837c/0043a5d6-b1fc-47aa-bad0-1a16e4b19c2c)

RomaricMourgues:
@rezk2ll is it fixed with the latest PR ? Can I put this again to QA column ?

hantt12:
Please fix thís. I will push thís backlog to TODO column https://github.com/linagora/Twake/issues/2525

RomaricMourgues:
Ok for issue #2525, I mark this one as qa:passed as it seems we are only talking about the #2525 now. (#2525 will be tagged canary:failed)

PR #2452, opened 2022-08-10
🌟 message delivery status (#2451)

rezk2ll:
## Description

- adds message delivery status
- adds an endpoint to mark messages as delivered
- adds an endpoint to mark messages as seen 

## Related Issue

#2451 

## Motivation and Context

- adds the ability to know if messages have been delivered, read

## How Has This Been Tested?

- locally

### being sent
image

### sent message
image

### failed message
image

### delivered message
image

### seen message
image

https://user-images.githubusercontent.com/6764881/185974730-92c1622f-f78e-4a63-9716-1eb465233d63.mp4

## Types of changes


RomaricMourgues:
Hi, that's impressive in the video ! I see that the icon goes from sent to delivered, then again to sent then to delivered again, why is happening here ?

RomaricMourgues:
Also could you update the documentation here: https://www.notion.so/linagora/API-Rest-Websockets-1f3f54f110b44743a102baf269f66ebd for the notification aknowledgement and create an issue for the mobile team to implement a call to this endpoint in background process when receiving a push notification ?

Then there is some documentation about marking messages as read here https://www.notion.so/linagora/Rest-APIs-Views-bookmarks-search-b022d821bb2a44babec096103c0fda2c. Same, mobile team should be informed via a new ticket.

And finally two points to be discussed:
- how does it behave inside threads
- what about the modal with the list of who read a specific message

rezk2ll:
> Hi, that's impressive in the video ! I see that the icon goes from sent to delivered, then again to sent then to delivered again, why is happening here ?

it was an issue with message propagation and WebSocket updates, fixed in https://github.com/linagora/Twake/pull/2452/commits/a97a9bdfed387435a2c4ca72f8782991b7232d06

rezk2ll:
> 



> Also could you update the documentation here: https://www.notion.so/linagora/API-Rest-Websockets-1f3f54f110b44743a102baf269f66ebd for the notification aknowledgement and create an issue for the mobile team to implement a call to this endpoint in background process when receiving a push notification ?
> 
> Then there is some documentation about marking messages as read here https://www.notion.so/linagora/Rest-APIs-Views-bookmarks-search-b022d821bb2a44babec096103c0fda2c. Same, mobile team should be informed via a new ticket.
> 
> And finally two points to be discussed:
> 
> * how does it behave inside threads
> * what about the modal with the list of who read a specific message

threads and the message display are gone when we introduced the quote reply feature.

RomaricMourgues:
> > 
> 
> > Also could you update the documentation here: https://www.notion.so/linagora/API-Rest-Websockets-1f3f54f110b44743a102baf269f66ebd for the notification aknowledgement and create an issue for the mobile team to implement a call to this endpoint in background process when receiving a push notification ?
> > Then there is some documentation about marking messages as read here https://www.notion.so/linagora/Rest-APIs-Views-bookmarks-search-b022d821bb2a44babec096103c0fda2c. Same, mobile team should be informed via a new ticket.
> > And finally two points to be discussed:
> > 
> > * how does it behave inside threads
> > * what about the modal with the list of who read a specific message
> 
> threads and the message display are gone when we introduced the quote reply feature.

Threads are still here in the non direct channel views no ?

And I was talking about this modal, with the UserReadCursors we should be able to compute this view what do you think? 
Screenshot 2022-08-29 at 15 31 42

PR #2467, opened 2022-08-30
🧹 cleaned unused imports, variables and fixed types

rezk2ll:


RomaricMourgues:
@rezk2ll why is it failing ?

rezk2ll:
> @rezk2ll why is it failing ?

fixed, it was caused by a removed side effect in the `useCurrentUser` hook.
```js
  //Depreciated way to get use update from LoginService
  LoginService.recoilUpdateUser = setUser;
```

Issue #2472, opened 2022-08-31
improve message delivery status


rezk2ll:
- enable message delivery status for all channels
- add the ability to see who read a specific message

PR #2474, opened 2022-08-31
🌟 Added message delivery status to all channels and added seen by UI

rezk2ll:
## Description

- enabled message delivery status for all types of channels
- added an endpoint to list the users who've seen a specific message

## Related Issue

#2472 

## Motivation and Context

- adds message delivery status to all channels
- adds the ability to detect who did read a message 

## How Has This Been Tested?

- locally

## Types of changes

### DIFF
--- a/twake/backend/node/src/services/channels/services/member/service.ts
+++ b/twake/backend/node/src/services/channels/services/member/service.ts
@@ -795,11 +795,15 @@ export class MemberServiceImpl {
    */
   async setChannelMemberReadSections(
     section: { start: string; end: string },
-    context: CompanyExecutionContext & { channel_id: string },
+    context: CompanyExecutionContext & { channel_id: string; workspace_id: string },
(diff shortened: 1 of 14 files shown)


twake/backend/node/src/services/channels/services/member/service.ts
```diff
@@ -874,4 +878,29 @@ export class MemberServiceImpl {
...
+   * @param {String} id - the message id
+   * @param {ChannelExecutionContext} context - the thread execution context
+   * @returns { Promise<string[]>} - the promise containing the user id list
+   */
+  async getChannelMessageSeenByUsers(
+    id: string,
+    context: ChannelExecutionContext,
+  ): Promise<string[]> {
+    const channelReadSections = await this.getChannelMembersReadSections(context);
+
```
RomaricMourgues, line 894: I propose we use the channel-stats “member_count” value to refuse executing this function if count is > 100.

Issue #2521, opened 2022-09-20
Migration tool to set every old messages as seen by everybody (set the cursor of everymembers)


RomaricMourgues:
Related to this comment: https://github.com/linagora/Twake/issues/2487#issuecomment-1251938703

PR #2534, opened 2022-09-22
🌟 added migration command to set every old messages as seen by everyone in public channels

rezk2ll:
## Description

- Added migration command to fix channel members read sections.

## Related Issue

- #2521 

## Motivation and Context

- old messages are not marked as seen by anyone.

## How Has This Been Tested?

- locally

## Types of changes


RomaricMourgues:
Note: even tho it doesn't work on private or direct channels because the user in the context is not a member of those channels so it can't fetch the members ( this check was made in the members service ).
and another thing, when the command finishes it doesn't give back the shell propmt somehow ( it doesn't indicate it finished after the spinner is done )

Issue #2622, opened 2022-11-25
Twake Drive to Node


RomaricMourgues:
## Database

```
# DriveFile, PK is [company_id, id] a secondary key is created over [company_id, parent_id, id]
{
    "id": string;
    "company_id": string; //New field

    "parent_id": string; // (empty string for root folder)
    "isintrash": boolean;
    "isdirectory": string;
    "name": string;
    "extension": string;
    "description": string;
    "tags": [];

    "added": string;
    "last_modified": string;
    "last_version_cache": typeof FileVersion; //New field

    //Access
    "access_info": {
        "public_access_token": string; //If set then it is possible to access this drive / folder using this token + IT IS ALWAYS A READONLY ACCESS
        "authorized_entities": { // If empty list, look at parent for access information
           "type": "user",
           "id": string
          }[];
        "unauthorized_entities": { // Has priority over authorized_entities but not on public_access_token
           "type": "user",
           "id": string
          }[];
    };

    //Other
    "content_keywords": string;
    "hidden_data": {};

    //Related to old Drive
    "workspace_id": string;
    "public_access_key": string;
    "root_group_folder": string;
    "creator": string;
    "size": number;
    "detached_file": boolean;
    "has_preview": boolean;
    "shared": boolean;
    "url": string;
    "preview_link": string;
    "object_link_cache": string;
    "external_storage": boolean;
    "last_user": string;
    "attachements": [];
}
```

```
# FileVersion
{
    //Id of the version of the file
    "id": string;

    //The file itself, using the existing new node "file" entity
    "provider": string | "drive" | "internal"; //Equivalent to "source" in twake/backend/node/src/services/messages/entities/message-files.ts
    "file_id": string;
    "file_metadata": same type than MessageFileMetadata; //New field

    "date_added": number;
    "creator_id": string;
    "application_id": string;

    //Related to previous Drive
    "realname": string;
    "key": string;
    "mode": "OpenSSL-2";
    "file_size": number;
    "filename": string;
    "data": {};
}
```

## Endpoints (WIP)

```
GET /internal/services/documents/v1/companies/:company_id/item/:id?public_access_token={optional_public_access_token}
> Returns an object with the requested DriveFile and the list of children and test access of current user to all of this item(s)
> If ":id" is empty, then return everything from root folder
> If ":id" is "trash", then return everything from root folder and where isintrash is true
```

```
DELETE /internal/services/documents/v1/companies/:company_id/item/:id
> Test access to item
[...]

PR #2632, opened 2022-11-29
📁 Add Drive backend in node #2622

rezk2ll:


twake/backend/node/src/services/documents/entities/drive-file.ts
```diff
@@ -0,0 +1,133 @@
...
+  @Column("id", "uuid", { generator: "uuid" })
+  id: string;
+
+  @Type(() => String)
+  @Column("parent_id", "uuid")
+  parent_id: string;
+
+  @Type(() => Boolean)
+  @Column("is_intrash", "boolean")
+  is_intrash: boolean;
```
RomaricMourgues, line 26: `is_in_trash` better than `is_intrash`

PR #2645, opened 2022-12-06
🤖 Kg changes

RomaricMourgues:
#2629 

We open an endpoint at:
`POST https://api.twake.app/internal/services/knowledge-graph/v1/push`

Authorised by a Token authorization header:
`Authorization: Token {some token defined together}`

And with the following data in JSON:
```
{
  events: [KnowledgeGraphCallbackEvent, KnowledgeGraphCallbackEvent, KnowledgeGraphCallbackEvent, ...]
}

type KnowledgeGraphCallbackEvent = {
  recipients: {
    type: "user";
    id: string; // KG user id which is a md5 of the email
  }[];
  event: {
    type: "user_tags"; //More events will be added later
    data: {
      //For user_tags event only
      tags?: {
        value: string;
        weight: number;
      }[];
    };
  };
};

```

The reply will be if everything was alright:
```
{
  "status": "success"
}
```

### DIFF
--- a/Documentation/docs/internal-documentation/backend-services/knowledge-graph/README.md
+++ b/Documentation/docs/internal-documentation/backend-services/knowledge-graph/README.md
@@ -1,0 +1,39 @@
+The knowledge graph can send us event at the following endpoint:
+`POST https://api.twake.app/internal/services/knowledge-graph/v1/push`
+
+Authorized by a Token authorization header:
+`Authorization: Token {some token defined together}`
+
+And with the following data in JSON:
+
+```
+{
+  events: [KnowledgeGraphCallbackEvent, KnowledgeGraphCallbackEvent, KnowledgeGraphCallbackEvent, ...]
+}
+
+type KnowledgeGraphCallbackEvent = {
+  recipients: {
+    type: "user";
+    id: string; // KG user id which is a md5 of the email
+  }[];
+  event: {
+    type: "user_tags"; //More events will be added later
+    data: {
+      //For user_tags event only
+      tags?: {
+        value: string;
+        weight: number;
+      }[];
+    };
+  };
+};
+
+```
+
+The reply will be if everything was alright:
+
+```
+{
+  "status": "success"
+}
+```
--- a/twake/backend/node/config/default.json
+++ b/twake/backend/node/config/default.json
@@ -145,6 +145,7 @@
   },
   "knowledge-graph": {
     "endpoint": "http://host-gateway:8888",
+    "callback_token": "secret",
     "use": false,
     "forwarded_companies": []
   },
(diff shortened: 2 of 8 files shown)


aioun:
Could you please confirm if this example will work for you. Different topics for each user and each topic has a weight.   Moreover, do you think that this will work for thousands of users ? Do you plan to handle the situation where a user has not any topic i.e.; the data is empty or when the user didn't receive yet any information about the topics. Thanks for all these!
{
  "events": [
    {
      "recipients": {
        "type": "user",
        "id": "00000000000000000000000000000"
      },
      "event": {
        "type": "topic",
        "data": [
           {
              "value": "projet collaboratif",
              "weight": 60
            },
            {
              "value": "Réunion du Lundi",
              "weight": 20
            }                  
        ]
      }
    },
     {
      "recipients": {
        "type": "user",
        "id": "11111111111111111111111111111111"
      },
      "event": {
        "type": "topic",
        "data": [
           {
              "value": "fonction javascript",
              "weight": 90
            },
            {
              "value": "Réunion du Lundi",
              "weight": 40
            }                  
        ]
      }
    }
  ]
}

RomaricMourgues:
- I would prefer to have the type "topics" or "user_topics"
- For the data, in case we have more to share than only topics, I would prefer also to have an additional nested value (event.data.topics = [] instead of event.data = [])

For the number of users yes, thousand will be the maximum, but preferably send it per pages of 500 users. And no problem for the 0-topics cases.

For the number of topics, please do not send more than 100 topics.

Also please normalise the weight between users, like a weight=40 in user1 and a weight=80 in user2 should mean user2 is twice more in this topic that user1 (or tell me otherwise).

Finally, please normalise the "values", to not have like "Fonction JavaScript", "fonctions javascript", "fonctions js" etc but a single entry string.

aioun:
Is this OK for your. You can use **type** to determine the nature of user data and update your interface accordingly. 
```json
{
  "events": [
    {
      "recipients": {
        "type": "user",
        "id": "00000000000000000000000000000"
      },
      "event": {
        "type": "user_topics",
        "data": [
           {
              "value": "projet collaboratif",
              "weight": 60  
            },
            {
              "value": "Réunion du Lundi",
              "weight": 20
            }                  
        ]
      }
    },
     {
      "recipients": {
        "type": "user",
        "id": "11111111111111111111111111111111"
      },
      "event": {
        "type": "user_style",
        "data": [
           {
              "value": "style",
              "tag1": "parameter_style"
            },
            {
              "value": "style",
              "tag1": "parameter_style"
            }                  
        ]
      }
    }
  ]
}
```

RomaricMourgues:
OK

RomaricMourgues:
I don’t know yet when our current sprint about Drive has ended On 8 Dec 2022, at 11:39, aioun ***@***.***> wrote:﻿
Thanks, when you think is possible to make a test ( sending only a Post as a first attempt ) and see the rendering on Twake ?

PR #2667, opened 2022-12-20
added drive access check

rezk2ll:


twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -514,18 +516,30 @@ export class DocumentsService {
...
-      ) {
-        return false;
+      if (token) {
+        if (!item.access_info.public.token) {
+          return false;
+        }
+
+        const { token: itemToken, level: itemLevel } = item.access_info.public;
+
+        return itemLevel === level && itemToken === token;
```
RomaricMourgues, line 526: Careful here, implicitelly if you have write access you also have read access:
"manage" includes "write" that includes "read" that includes "none".
Better to have a little utility for that, you could simply do this:
```
manage: 30
write: 20
read: 10
none: 0

hasLevel(level, requested) => level >= requested
```

twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -514,18 +516,30 @@ export class DocumentsService {
...
       }
 
       if (
-        item.access_info.unauthorized_entities.find(entity =>
-          entityIdentityCheck(entity.id, context),
-        )
+        !item.access_info.entities.find(entity => {
+          if (entity.level !== level) return false;
```
RomaricMourgues, line 531: Same than previous comment

twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -514,18 +516,30 @@ export class DocumentsService {
...
+
+          if (entity.type === "user" && entity.id === context.user.id) {
+            return true;
+          }
+
+          if (entity.type === "company" && entity.id === context.company.id) {
```
RomaricMourgues, line 537: Here we will also check that the user is not a `guest` I did a similar function in this same file to test for admins. So user should be at least company member (admin, owner also working)

twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -538,23 +552,7 @@ export class DocumentsService {
...
-        if (
-          !child.access_info.authorized_entities.find(entity =>
-            entityIdentityCheck(entity.id, context),
-          )
-        ) {
-          return false;
-        }
-
-        return true;
+        return this.checkAccess(child.id, child, level, context, token);
```
RomaricMourgues, line 555: This part is wrong, we do not prevent access to a folder because we don't have access to one of the children, we'll cover this case in the "get" method (already done) so here we should `return true`

twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -514,18 +516,30 @@ export class DocumentsService {
...
+          if (entity.level !== level) return false;
+
+          if (entity.type === "user" && entity.id === context.user.id) {
+            return true;
+          }
+
+          if (entity.type === "company" && entity.id === context.company.id) {
+            return true;
+          }
+
```
RomaricMourgues, line 540: Something is missing here, the "folder" case it looks like this:
```
if(entity.type === "folder" && entity.id !== this.ROOT && entity.id !== this.TRASH && await checkAccess(entity.id, ...)){
  return true;
}
```

To avoid starting with this expensive line, maybe do it as a last step or order the auth_entities to test this folder case in last position.

RomaricMourgues (changes requested):

twake/backend/node/src/services/documents/utils.ts
```diff
@@ -105,3 +113,241 @@ const generateAccessToken = (): string => {
...
+    }
+
+    if (token) {
+      if (!item.access_info.public.token) {
+        return false;
+      }
+
+      const { token: itemToken, level: itemLevel } = item.access_info.public;
+
+      return hasAccessLevel(itemLevel, level) && itemToken === token;
```
RomaricMourgues, line 324: Return here only if hasAccessLevel is true, if not we go to the other tests

twake/backend/node/src/services/documents/utils.ts
```diff
@@ -105,3 +113,241 @@ const generateAccessToken = (): string => {
...
+    }
+
+    return !!item.access_info.entities.find(async entity => {
```
RomaricMourgues, line 327: The folder test must comes last, so add some sort here to make sure we end with the channel test and then the folder test.

twake/backend/node/src/services/documents/utils.ts
```diff
@@ -105,3 +113,241 @@ const generateAccessToken = (): string => {
...
+        case "user":
+          return entity.id === context.user.id;
+
+        case "company":
+          if (entity.id === context.company.id) {
+            return !isCompanyGuest(context);
+          }
+
+          return false;
+        case "folder":
```
RomaricMourgues, line 340: Add the case "channel" where we'll test that the user is member of that channel, for now we return false always (until I implement it)

RomaricMourgues (changes requested):
Very small changes, thanks for everything else it is perfect 👍

PR #2677, opened 2023-01-03
📁 #2622 drive enhancements

rezk2ll:


RomaricMourgues (changes requested):
Good to me, please just change the documentation to not mention the old fields that we don't use anymore (see initial issue comment)

rezk2ll:
> Good to me, please just change the documentation to not mention the old fields that we don't use anymore (see initial issue comment)

done

Issue #2652, opened 2022-12-08
Drive Node Additional Features


RomaricMourgues:
- [x] ZIP download: Send a list of ids (files and/or folders) and automatically starts downloading a zip. The zip will be constructed while being downloaded (no wait time to build the zip), we did in in PHP so should be OK in node.

- [x] Search: In three steps:
       1. Extract during the thumbnail generation time the important words of the documents (max 250 unique words ordered by lower frequency and term length first). And save the result to the content_keywords field.
       2. Create a `drive-file.search.ts` and alter `drive-file.ts` to implement the search object (see user.ts and user.search.ts for example)
       3. Create the `POST /companies/:company_id/search` to search on multiple fields (tags, name&content, creation date, creator)

- [x] Auto rename folders and files with same name in the same folder after creating / updating / moving a document

- [x] Prevent moving a folder inside itself or one of its children

- [x] Do not trust the frontend data and get the drive item size from the version, and also get the up to date version from the file service (so the frontend in fact should just send you some id and the rest is computed in backend)

- [x] Size computation of parents not working

- [x] Cleanup of entities (remove unused fields relative to old php entities)

- [x] **(done by Romaric)** Realtime endpoint, publish any change to an item to the realtime endpoint corresponding to the parent of this document.

PR #2679, opened 2023-01-04
📁 #2652 added download drive item as a zip archive

rezk2ll:
related to #2652


RomaricMourgues (commented):
It seems to be exactly what I had in mind 👍 
- Could you check what happen in client closes the connection during the zip generation ?
- Also what happens for files that cannot be downloaded (links, external files etc), we should ignore them for now

RomaricMourgues (approved):

RomaricMourgues:
All good to me

rezk2ll:
> I'll merge when you want me to merge

you can merge it.

RomaricMourgues:
Forgot about it, thanks

PR #2685, opened 2023-01-10
📁 Added Drive file search

rezk2ll:


RomaricMourgues (commented):
Could we use better names than finish-service and start-service ?
Like "extract-keywords" and "save-keywords-to-search"

PR #2715, opened 2023-02-01
📁 Fixed drive item versions not working with Cassandra

rezk2ll:
## Description

- changed the order of the primary key fields to fix the drive item versions fetching when using the Cassandra database.

## Related Issue

#2714

## Motivation and Context

- fixes versions not showing up in the drive.

## How Has This Been Tested?

- locally

## Types of changes

### DIFF
--- a/twake/backend/node/src/services/documents/entities/file-version.ts
+++ b/twake/backend/node/src/services/documents/entities/file-version.ts
@@ -4,7 +4,7 @@ import { Column, Entity } from "../../../core/platform/services/database/service
 export const TYPE = "drive_file_versions";
 
 @Entity(TYPE, {
-  primaryKey: [["file_id"], "id"],
+  primaryKey: [["drive_item_id"], "id"],
   type: TYPE,
 })
 export class FileVersion {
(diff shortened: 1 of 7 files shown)


twake/backend/node/src/services/documents/entities/file-version.ts
```diff
@@ -4,7 +4,7 @@ import { Column, Entity } from "../../../core/platform/services/database/service
 export const TYPE = "drive_file_versions";
 
 @Entity(TYPE, {
-  primaryKey: [["file_id"], "id"],
+  primaryKey: [["id"], "file_id"],
   type: TYPE,
 })
 export class FileVersion {
```
RomaricMourgues, line 10: Please rename "id" to "drive_file_id", I think it will be more clear wdyt ? @rezk2ll

rezk2ll: I think the `file_id` is supposed to be `drive_file_id` and the `id` is just the id of the version

RomaricMourgues: In this case there is still an issue because when retrieving the list of versions you do it per "drive_file_id", it is like a folder structure: `/{drive_file_id}/{version_id}` so you can get all the versions from a drive_file_id or a single version but then need both drive_file_id and version_id.
In the change you made the directory structure is `/{id}/{file_id}` so file_id can't be equivalent to the drive_file_id right ?

RomaricMourgues: @rezk2ll I let you see with @MontaGhanmy

Issue #2709, opened 2023-01-31
[Drive] Copy public link not working


MontaGhanmy:
**Describe the bug**
- File -> Access management -> Copy public link to clipboard, not working, nothing in clip board when testing in staging
- "Copy public link” should be one of the options in the item drop down in case the item is publicly shared.

PR #2719, opened 2023-02-01
📁 Fix copy public link #2709

MontaGhanmy:
## Description
File -> Access management -> Copy public link to the clipboard, not working, nothing in clipboard when testing in staging.

## Motivation and Context
The old document.execCommand('copy') method is used to execute a "copy" command on the current document. This method has been implemented in various web browsers, but the exact behavior and implementation may vary between different browsers. In some cases, the document.execCommand('copy') method may not work as expected.

## Types of changes

### DIFF
--- a/twake/frontend/src/app/features/global/utils/CopyClipboard.ts
+++ b/twake/frontend/src/app/features/global/utils/CopyClipboard.ts
@@ -1,8 +1,13 @@
-export function copyToClipboard(url: string): void {
-    const el = document.createElement('textarea');
-    el.value = url;
-    document.body.appendChild(el);
-    el.select();
-    document.execCommand('copy');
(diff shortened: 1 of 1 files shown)


RomaricMourgues (approved):

twake/frontend/src/app/features/global/utils/CopyClipboard.ts
```diff
@@ -1,8 +1,13 @@
...
+    } catch (err) {
+      const el = document.createElement('textarea');
+      el.value = url;
+      document.body.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      document.body.removeChild(el);
+    }
+  }
+
```
rezk2ll, line 13:
```suggestion
}
```

PR #2726, opened 2023-02-06
📁 Fixed downloading folder / mix of folders and documents not working

MontaGhanmy:
## Types of changes


twake/backend/node/src/services/documents/services/index.ts
```diff
@@ -653,22 +653,37 @@ export class DocumentsService {
...
         try {
-          await addDriveItemToArchive(id, null, archive, this.repository, context);
+          counter = await addDriveItemToArchive(
+            id,
+            null,
+            archive,
+            this.repository,
+            context,
+            undefined,
+            counter,
```
rezk2ll, line 674:
```suggestion
            counter,
```

```suggestion
            undefined,
```

probably we can swap the prefix and counter params order since counter is always passed and the prefix is optional.
something like:
```
export const addDriveItemToArchive = async (
  id: string,
  entity: DriveFile | null,
  archive: archiver.Archiver,
  repository: Repository<DriveFile>,
  context: CompanyExecutionContext,
  counter: number,
  prefix?: string,
): Promise<number> => {
```

PR #2731, opened 2023-02-07
added Drive backend e2e tests

rezk2ll:
image


RomaricMourgues:
Nice ! I wait for the CI and then I will merge it. We'll get back to it when test coverage is visible somewhere thanks to @MontaGhanmy :)

RomaricMourgues:
I think it fails because some assets was not added to the git :)

rezk2ll:
> I think it fails because some assets was not added to the git :)

yes the file was empty, fixed

PR #2751, opened 2023-02-17
📁 Search drive files

MontaGhanmy:


twake/frontend/src/app/components/search-popup/tabs/drive.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+export const DriveItemsResults = (props: { max?: number }) => {
+  const { driveItems, loading } = useSearchDriveItems();
+
+  if (driveItems.length === 0 && !loading) return <NothingFound />;
+
+  return (
+    <>
+      {driveItems.slice(0, props?.max || driveItems.length).map(item => (
+        <FileResult key={item.id} file={item} />
```
RomaricMourgues, line 39: Does this work when clicking on it ? Does it redirect to the Drive correct location ?

twake/frontend/src/app/features/drive/api-client/api-client.ts
```diff
@@ -86,4 +103,20 @@ export class DriveApiClient {
         appendPublicToken(true),
     );
   }
+
+  static async search(searchString: string, options?: BaseSearchOptions) {
+    const companyId = options?.company_id ? options.company_id : Workspace.currentGroupId;
+    const query = `/internal/services/documents/v1/companies/${companyId}/search`;
```
RomaricMourgues, line 109: Did you check that we can't access restricted results through search ? I think on the backend side we had to add some access check.

RomaricMourgues (changes requested):
Some questions :)

twake/frontend/src/app/components/search-popup/tabs/drive.tsx
MontaGhanmy: I just updated the entire component for displaying the result item and the download button is working now. 

Screenshot 2023-02-21 at 10 39 40 AM

RomaricMourgues:
We also need a button to "jump" to the right folder in the Drive 🙈 Could you do that ? :p

MontaGhanmy:
> We also need a button to "jump" to the right folder in the Drive 🙈 Could you do that ? :p

Yes we can :)

RomaricMourgues:
@MontaGhanmy ready to merge ?

Issue #2752, opened 2023-02-20
[Drive] Fix breadcrumb navigation ui issues


rezk2ll:
![image](https://user-images.githubusercontent.com/6764881/220054701-04b89f24-7fa6-4293-9b6b-f49a8eff0437.png)

PR #2753, opened 2023-02-21
📁 #2752 fixed breadcrumb navigation overflow

rezk2ll:
## Description

- changed the style of the breadcrumb navigation to remove the overflow and wrap the elements.

## Related Issue

#2752 

## Motivation and Context

solves a UI issue with the breadcrumb navigation having an overflow:

image
 

## How Has This Been Tested?

- locally

#### while connected:

image

#### incognito / shared folder:

image

## Types of changes

### DIFF
--- a/twake/frontend/src/app/views/applications/drive/header-path.tsx
+++ b/twake/frontend/src/app/views/applications/drive/header-path.tsx
@@ -32,7 +32,7 @@ export const PathRender = ({
   onClick: (id: string) => void;
 }) => {
   return (
-    <Title className="overflow-hidden mr-2 pl-px">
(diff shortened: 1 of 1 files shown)


RomaricMourgues:
Héhé if we do that the we have a new problem if there is too many folders no ?
The solution is probably to do this kind of "..." menu, but it's slower.

An `overflow-auto` should work in the meanwhile but only if the paddings / marings are OK (the overflow auto should not display scroll bar if the content isn't bigger that the container). So I think we should look at this issue instead of putting a overflow-hidden. What do you think ?

Screenshot 2023-02-21 at 10 43 10

rezk2ll:
@RomaricMourgues we can also make the middle elements shrink instead.

image

image

PR #2758, opened 2023-02-22
📁 Restrict access for drive search

MontaGhanmy:


RomaricMourgues (approved):
Not very optimised but good for v1

